PR #27, opened 2021-04-01
Add resample_raw

richardkoehler:
Implement MNEs resample function to downsample (or upsample) the raw data. This can be used to substantially speed up computation time or to homogenize sampling frequencies for group analyses. Simply set resample_raw to false if resampling is not desired.

See: [mne.filter.resample](https://mne.tools/stable/generated/mne.filter.resample.html#mne.filter.resample)


timonmerk:
Thanks for working on this @richardkoehler! 
Actually a very "simple" resample function is already implemented in test_features.py
```

    label_downsampled = dat_[:,
                        ::int(np.ceil(fs / settings["resampling_rate"]))]
```

It's also named in the settings. So it's a bit a question if we want to allow resampling from the module side as well and not from the amplifier.

richardkoehler:
@timonmerk  I implemented this, because the MNE function can downsample by non-integer steps, so it can downsample to any frequency and it is optimized for neural signals!

I had 2 different use cases in mind: 
1. Offline analysis, when you want to downsample for computational efficiency (e.g. running your analysis at 7500Hz sampling rate would take 10x as long as the recording duration) or to homogenize sampling rates across recordings before analysis.
2. Online analysis, if you want to hardware-sample at a high frequency (e.g. for simultaneously saving the data for later offline analysis), but need to downsample for real-time analysis because the hardware sampling rate is too high.

timonmerk:
Ok, agree. I think of a different name for settings.json "resampling_rate", since this sounds very similar. Maybe "resampling_rate_features" makes more sense?

PR #31, opened 2021-04-02
Adapted sharpwave features

### LINKED ISSUE #9: Add additional sharpwave features
Currently only prominence and sharpness are implemented. 

Optimize function by speed, and add features in "features_" returning dataframe:

- width
- interval
- decay time 
- rise time 
- rise steepness
- fall steepness 
- slope ratio

timonmerk:
@richardkoehler  if you want to look, I adapted the sharpwave estimation. It took me quite some time to get around this by not having the same idea how I programmed it previously :) 

The settings.json have now an option to enable the separate features, but also different estimators. 
This speeds up the sharpwave estimation by a factor of 10-20, by specifically turning off the sw features it should be faster. 

The test.py is adapted, but I have to think of a separate test to see if the features are correctly estimated still. 

Fixes #9

### DIFF
--- a/examples/settings.json
+++ b/examples/settings.json
@@ -52,10 +52,6 @@
     },
 
     "sharpwave_analysis_settings": {
-        "filter_low_cutoff" : 5,
-        "filter_high_cutoff" : 90,
-        "distance_troughs" : 5,
-        "distance_peaks" : 1,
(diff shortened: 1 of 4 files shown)


richardkoehler:
Wow looks great, can't wait to try them out!

PR #32, opened 2021-04-03
Change type

### LINKED ISSUE #7: save MNE Raw_Array Info in feature folder
Currently 

- settings.json 
- df_M1.p
- FEATURES.p

are saved in the run_path folder. The MNE Array Info files, which contains all BIDS information should be saved as well.

### LINKED ISSUE #24: Rewrite output to csv
pickle is python specific, csv files could be also viewed on github/at the folder

### LINKED ISSUE #29: Replace "ECOG" column in M1.tsv with "type"
I would suggest replacing the ECOG column with something like "type" and then specifying the type expicitly, e.g. "ecog", "dbs", "eeg", etc. 
This would give us much more flexibility in the future. Also, "subcortex" channels in the rereferencing function are inferred implicitly by picking channels that are "used"==1 but not "ECOG"==1. By having the actual channel type we could specify subcortex channels explicitly, by e.g. picking channels with the type "dbs"==1.

@timonmerk  what do you think?

timonmerk:
- tests run through
- types are set to mne.RawArray types, subcortex and cortex indices are easier made now 
- electrode position files are added to example BIDS data 
- electrode position and fs is saved after feature estimation to settings.json (not nice way, but saves an additional toolbox generating file, or having to reload BIDS data only for position information and fs) 

Fixes #29 
Fixes #24 
Fixes #7


examples/test_features.py
```diff
@@ -52,9 +52,9 @@ def est_features_run(PATH_RUN) -> None:
 
     to_ref_idx = np.array(df_M1[(df_M1['used'] == 1)].index)
 
-    cortex_idx = np.where(df_M1.ECOG == 1)[0]
-    subcortex_idx = np.array(df_M1[(df_M1["ECOG"] == 0) & (df_M1['used'] == 1)
-                                   & (df_M1['target'] == 0)].index)
+    cortex_idx = np.where(df_M1.type == 'ecog')[0]
+    subcortex_idx = np.array(df_M1[(df_M1["type"] == 'seeg') | (df_M1['type'] == 'dbs')
+                                   & (df_M1['type'] == 'lfp')].index)
```
richardkoehler, line 57: @timonmerk  Shouldn't there be a logical "or" before the "lfp" typecheck? Like this:
`| (df_M1['type'] == 'lfp')`

pyneuromodulation/tests/test.py
```diff
@@ -79,9 +79,9 @@ def initialize_rereference(df_M1):
 
     to_ref_idx = np.array(df_M1[(df_M1['used'] == 1)].index)
 
-    cortex_idx = np.where(df_M1.ECOG == 1)[0]
-    subcortex_idx = np.array(df_M1[(df_M1["ECOG"] == 0) & \
-                    (df_M1['used'] == 1) & (df_M1['target'] == 0)].index)
+    cortex_idx = np.where(df_M1.type == 'ecog')[0]
+    subcortex_idx = np.array(df_M1[(df_M1["type"] == 'seeg') | (df_M1['type'] == 'dbs')
+                                   & (df_M1['type'] == 'lfp')].index)
```
richardkoehler, line 84: @timonmerk  Same as above:
Shouldn't there be a logical "or" before the "lfp" typechecke like: 
`| (df_M1['type'] == 'lfp')`

richardkoehler:
@timonmerk  Nice commit! We probably also have to adapt the rereference Module accordingly then right? I can take care of that if you want.

richardkoehler (changes requested):
@timonmerk  I had two small issues, but not sure if these are correct :)

timonmerk:
@richardkoehler your comments were correct, actually you could also fix them inline in github by clicking the left "+" sign. This let's you automatically commit new changes also. 
Now added the changes by direct commit to main

PR #36, opened 2021-04-07
Update rereference and test_features

richardkoehler:
@timonmerk  So I was pulling all the information extraction from df_M1, like cortex_idx and subcortex_idx etc., into the rereference module, to make the actual analysis script (e.g. test_features) a bit cleaner. This got me thinking that we should maybe initialize the settings as a class, into which we load the settings.json and M1.tsv. We could then get the relevant information via functions e.g. ch_names(), used_chs(), cortex_idx(), out_path() etc. What do you think?
The pull request doesn't implement the class yet or anything, I only pulled the cortex_idx etc. into the rereference initialization.

### DIFF
--- a/examples/test_features.py
+++ b/examples/test_features.py
@@ -47,18 +47,7 @@ def est_features_run(PATH_RUN) -> None:
         if PATH_M1 is not None and os.path.isfile(PATH_M1) \
         else define_M1.set_M1(raw_arr.ch_names, raw_arr.get_channel_types())
 
-    ch_names = list(df_M1['name'])
-    refs = df_M1['rereference']
-
-    to_ref_idx = np.array(df_M1[(df_M1['used'] == 1)].index)
-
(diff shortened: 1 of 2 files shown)


timonmerk:
Thanks @richardkoehler. Simulataneously I restructured the file as well, so now there will be some merge conflicts that need to be resolved. 

I like the settings class idea, that makes the analysis much more clean. I will have a look. 

The thing is that everything will look very different if we actually run this in real time e.g. with LSL, so the functions need to be written in a very general manner.

richardkoehler:
@timonmerk  Okay, then we can also just see how the real-time analysis works out and then adapt the settings accordingly. Should I still merge the pull request? You can then resolve the merge conflicts afterwards, I didn't actually make any huge changes.

timonmerk:
@richardkoehler I think I will adapt it manually afterwards, I renamed and deleted the file, so I think Git won't be able to track it. But let's see, I will make a PR myself.

PR #37, opened 2021-04-07
Add lsl realtime

### LINKED ISSUE #34: Test Non-BIDS Feature Estimation
Write Module wrapper using LSL

### LINKED ISSUE #35: Test py_neuromodulation env

timonmerk:
Edited: 
 - LSL function that requires running pylsl (e.g. through openBCI)
 - ReadME.m adapted for setup using requirements.txt for conda virtual environment install 
 - different issues popped using using the adapted fs using openBCI (120 Hz)

Issues: 
 - we don't check the sampling frequency wrt to the used notch filters
 - mne.filter.create_filter might fail, we don't check for that
 - define_M1 should also take EEG type, not just ECoG 

Fixes #35 
Fixes #34


timonmerk:
@richardkoehler  something I also figured out, now I deleted the gamma bands, and in such way also the entries. Maybe it makes sense to specify here also a true / false? Not sure though, since ppl might change bands individually. 

I think it makes sense to merge this PR first, and then yours, since we would mainly adapt start_LSL + start_BIDS, which are now new files.

richardkoehler:
@timonmerk  This is great! Can't wait to try it out! Do you think LSL is also an option for experiments with the TMSi amplifier?

Issue #38, opened 2021-04-07
Check in settings.py sampling frequency filter relation 


timonmerk:
For too low sampling frequencies, the current notch filter will fail. 
Design filter in settings.py? 

Settings.py should be written as class s.t. settings.json, df_M1 and 
such code isn't part of start_BIDS / start_LSL anymore. 

```
    fs = int(np.ceil(raw_arr.info["sfreq"]))
    line_noise = int(raw_arr.info["line_freq"])

    # read df_M1 / create M1 if None specified
    df_M1 = pd.read_csv(PATH_M1, sep="\t") \
        if PATH_M1 is not None and os.path.isfile(PATH_M1) \
        else define_M1.set_M1(raw_arr.ch_names, raw_arr.get_channel_types())
    ch_names = list(df_M1['name'])
    refs = df_M1['rereference']
    to_ref_idx = np.array(df_M1[(df_M1['used'] == 1)].index)
    cortex_idx = np.where(df_M1.type == 'ecog')[0]
    subcortex_idx = np.array(df_M1[(df_M1["type"] == 'seeg') | (df_M1['type'] == 'dbs')
                                   | (df_M1['type'] == 'lfp')].index)
    ref_here = rereference.RT_rereference(ch_names, refs, to_ref_idx,
                                          cortex_idx, subcortex_idx,
                                          split_data=False)
```

timonmerk:
@richardkoehler I am referencing from the PR to this issue here. 
It's basically not just the notch filter, but also the filter in generell. For the notch filter it might be easy to fix, since we initialize the filter currently in this manner in features.py 

```
data = notch_filter(x=data, Fs=self.fs, trans_bandwidth=15,
                    freqs=arange(self.line_noise, 3*self.line_noise,
                                 self.line_noise),
                    fir_design='firwin', verbose=False,
                    notch_widths=3, filter_length=data.shape[1]-1)
```

So at least the "3*self.line_noise" needs to be adapted wrt to the used sampling frequency. 
Maybe the notch / bandpass filter could be defined even in settings.py? In this way we could pass them to features.py instantiation, and the correct filters could be already precalculated.

timonmerk:
Something more tricky is probably the 'trans_bandwidth', since this heavily depends on the 'filter_length'. Higher 'filter_length' will allow for "sharper" filter functions. So it might be to iteratively to increase this parameter until the filter calculation is possible, or a more advanced way it possbile :sweat_smile:

[...]

Issue #12, opened 2021-03-29
Add feature grid projection 


timonmerk:
Define cortical + subcortical grid. Features are then interpolated based on coordinates

PR #39, opened 2021-04-09
Fix filter

timonmerk:
This pr restructures the repo s.t. settings.py is used as a class that stores settings.json and df_M1. 
This requires to change most parameter handling, tests and feature result reading / writing. 

@richardkoehler feel free have a look. 

Added IO module for BIDS handling as well as label write. 

Tests run through. 

Next steps are basically #38 
I will now focus on #12

This toolbox actually changed my view about adding as many features as possible :) https://www.cell.com/cell-systems/fulltext/S2405-4712(17)30438-6?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS2405471217304386%3Fshowall%3Dtrue
Probably it's better to focus on the established ones and make them real time compatible.


richardkoehler:
@timonmerk  Wow that's a big push! Yes then let's definitely focus on the real-time aspect and the most important features. Notch-filter is currently done by mne and fails for low sampling frequencies is that right? How low exactly? So should we rewrite the notch-filtering step so that a notch filter is calculated according to the sampling frequency, before the analysis is started?

PR #55, opened 2021-04-18
Relative path

mousa-saeed:
added relative path.
path and model parsing from command line


timonmerk (changes requested):
Thanks @mousa-saeed, I do not see the changes here in 'sub-testsub_ses-EphysMedOff_task-buttonpress_ieeg_ML_RES.p'. 
I assume this file is not necessary for pr right? 
Maybe you can remove it? 

Otherwise looks great to me!

timonmerk:
@mousa-saeed  Could you also fix this in example_BIDS.py and example_read_features.py? 
I think that's a great way to circumvent private paths 👍

timonmerk:
Ok, let's adapt the sys.path.append(os.path.join(os.pardir,'pyneuromodulation')) in the other modules as well

mousa-saeed:
should i adopt the example_BIDS.py to the test data? 
as this line doesnt make sense with the repo test data
PATH_RUN = os.path.join(BIDS_EXAMPLE_PATH, 'sub-testsub', 'ses-EphysMedOff',
                            'ieeg', "sub-testsub_ses-EphysMedOff_task-buttonpress_ieeg.vhdr")

PR #56, opened 2021-04-18
Relative path

timonmerk:
Reopen pr, since I only closed the recent one and didn't merge


timonmerk:
We should not forget to call pyreqs, to push a updated version of requirements.txt

PR #90, opened 2021-05-14
bug fix

timonmerk:
@richardkoehler I think ther's a bug in features.py
'est_connect' is estimated every time, even though it might not be enabled


richardkoehler:
Oops... 😅

Issue #95, opened 2021-08-12
restructure files and make py_neuromodulation a package


timonmerk:
create __init__.py file

Issue #94, opened 2021-08-12
Create env.yml file for environment setup


timonmerk:
will be faster and cleaner setup

PR #100, opened 2021-08-12
restructuring

timonmerk:


timonmerk:
Fixes #97, #96, #95 and #94

PR #115, opened 2021-11-02
PR Modularization of CohortWrapper

timonmerk:
- adapt settings.json for different fdomain features
 - add detection rate estimation
 - refactoring performance_out dict write up


timonmerk:
@richardkoehler Sorry again, I should adapt my changes more frequently in main!

PR #137, opened 2021-12-28
Optimize speed of feature normalization

### LINKED ISSUE #136: Speed up feature normalization
Feature normalization currently takes a lot of time, as it is implemented using pandas Series and DataFrames.

Proposed changes:
- Implement feature normalization using numpy arrays to speed up computation time
- Reduce memory usage by only storing feature absolutely needed for feature normalization

richardkoehler:
- Computation time of feature normalization was improved
- Memory usage of feature normalization was improved
- nm_eval_timing was repaired, as these changes broke the api

closes #136

### DIFF
--- a/py_neuromodulation/nm_eval_timing.py
+++ b/py_neuromodulation/nm_eval_timing.py
@@ -163,6 +163,6 @@ class NM_Timer:
 
         print("Average duration per function:")
         for key, val in dict_timings.items():
-            if key == "time_sw":
-                print("Duration per channel:")
             print(f"  {key} : {np.round(val*1000, 2)}ms")
+
+        print("fft, sw, bandpass, coherence and stft are timings for an individual channel")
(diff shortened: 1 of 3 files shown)


timonmerk:
Many thanks @richardkoehler, it works really well. 
I added the norm method as an enum implementation to make it a bit cleaner. What do you think?

It also does not take longer:
  time_rereference : 2.31ms
  time_resample : 2.92ms
  time_notchfilter : 20.62ms
  time_feature_norm : 0.29ms
  time_fft : 0.39ms

richardkoehler:
@timonmerk  I like it, it makes the code a lot cleaner! However, now this got me thinking that we should probably convert the normalization process into an Object as well (like you did with the other preprocessing steps). We could define a Normalizer ABC with FeatureNormalizer and RawNormalizer and then pass these Instances to nm_run_analysis.

timonmerk:
@richardkoehler That's true, but right now I am pretty happy with the setup, since the timing is optimized and the overall function setup is also quite clean. So maybe a non pending issue for future setups.

PR #143, opened 2022-01-06
Bay opt example

timonmerk:


examples/example_BIDS.py
```diff
@@ -18,9 +20,10 @@ def run_example_BIDS():
     nm_BIDS = nm_BidsStream.BidsStream(PATH_RUN=PATH_RUN,
                                        PATH_BIDS=PATH_BIDS,
                                        PATH_OUT=PATH_OUT,
-                                       LIMIT_DATA=False)
+                                       LIMIT_DATA=False,
+                                       VERBOSE=True)
 
-    nm_BIDS.run_bids()
+    #nm_BIDS.run_bids()
```
richardkoehler, line 26: @timonmerk Was this comment on purpose?

timonmerk: Ohh, I forgot uncommenting it, since I tested only the bayesian opt. code.  But I will undo it

PR #145, opened 2022-01-09
Fix pathing problem in nm_analysis

richardkoehler:
The current code only worked if you passed the filename with the suffix ".vhdr" because of the hardcoded `-len(".vhdr")`. I got an error because of only passing the files basename. My proposed fix should now work in both cases.
On a note, I think it is not necessarily recommended to use "import _pickle", because the standard module "import pickle" automatically does just this (https://stackoverflow.com/questions/19191859/what-difference-between-pickle-and-pickle-in-python-3).

Changes:
  - pass responsability for retrieving file stem to Path()
  - add missing comma after "SQUARED_ROTAWHEEL"
  - sort imports

### DIFF
--- a/py_neuromodulation/nm_analysis.py
+++ b/py_neuromodulation/nm_analysis.py
@@ -1,19 +1,21 @@
 import os
+from pathlib import Path
 from re import VERBOSE
-import pandas as pd
-import numpy as np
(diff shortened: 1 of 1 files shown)


timonmerk (approved):
Nice!

PR #147, opened 2022-01-09
Fix ClassMissingException

richardkoehler:
ClassMissingException previously printed the message "only one class present" even when the module nm_decode was simply initialized. Now ClassMissingException only prints this message when an Exception is actually raised.
Additionally, an individual message can be passed upon raising the Error, to potentially discern where exactly in the module the Error was raised.


timonmerk (approved):
Thanks for noting!

PR #148, opened 2022-01-09
Bugfix RT_rereference

richardkoehler:
- A bug was present where channels were used even when used was 0
  -> Now indexes are split into "*_used" and "*_good"
- Duplicates in the docstring were removed
- The separator for a multiple-channel reference is now "&" instead of "+"
  - "+" may also be used for DBS channels, e.g. "LFP_2+3+4_STN_BS"
  - if you want a multiple-channel reference you can now write "LFP_2+3+4_STN_BS&LFP_5+6+7_STN_BS"

### DIFF
--- a/py_neuromodulation/nm_rereference.py
+++ b/py_neuromodulation/nm_rereference.py
@@ -1,46 +1,40 @@
 from numpy import empty_like, mean, ndarray, where
-from pandas import isnull
+import pandas as pd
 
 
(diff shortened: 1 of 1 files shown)


timonmerk (approved):
Thanks for noting @richardkoehler! Somehow the "used" column was never tested it seems.
Also I could quickly verify using tests/test_rereference.py that the code still works :)

PR #149, opened 2022-01-09
Fix fs in SIDECAR.json

richardkoehler:
- sampling freq "fs" written out was the original freq, not the final freq
- SIDECAR.json now includes:
   - "fs"
   - "original_fs"
   - "ch_names" (channel names used to calculate features)
- Small typo in "_set_notch_filter" fixed
- Type hinting for "fs" fixed (can be float or int)

### DIFF
--- a/py_neuromodulation/nm_IO.py
+++ b/py_neuromodulation/nm_IO.py
@@ -43,7 +43,7 @@ def read_BIDS_data(PATH_RUN, BIDS_PATH):
     return (
         raw_arr,
         raw_arr.get_data(),
-        int(np.ceil(raw_arr.info["sfreq"])),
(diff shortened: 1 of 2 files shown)


timonmerk:
@richardkoehler Makes a lot of sense to write out both sampling frequencies!

In nm_bids.py `read_BIDS_data` the sampling frequency is casted on the other hand to int: `int(np.ceil(raw_arr.info["sfreq"]))`.

So I removed that, since declaring the typehint to float does not make that much sense if it would be an int anyway.
Also I tested running the code through, and in nm_generator `fs` is then converted to int at later stages when it actually becomes necessary for np slicing.

timonmerk (approved):

PR #151, opened 2022-01-10
Add test for example run

timonmerk:


timonmerk:
Thanks @richardkoehler for the conda install template 👍 
@jgvhabets The ReadMe.rst explains that with the `conda develop .` command you should be able to install the package in the current environment. Afterwards it should be able to import it from anywhere with that activated environment.

jgvhabets:
thanks @richardkoehler , it also worked for me:
 just tried to run import py_nmd from another folder (which did not work before). I ran conda develop first in my terminal, got a error that a path is required, but afterwards I am able to import py_neuromdoulation functions.
I just closed the conda-env, visual code etc, and I can still import from another directory.

PR #156, opened 2022-01-19
Fix cohort analysis

timonmerk:


timonmerk (commented):
@richardkoehler Sorry for this overblown PR.
FYI, I fixed a bug here in nm_normalization for the first sample:
1. `current` should not be zeros in `normalize_features`
2. for `normalize_raw` I stick with the same idea and return `current` twice

Returning zeros would even results in `NaN's` in some cases, which in turn force features to become zero.

timonmerk:
It's the commit "fix normalization for first sample"

PR #162, opened 2022-02-11
Refactor nm_define_nmchannels

richardkoehler:
Changes:
- No functionality was changed
- Type hints added
- Code cleaned up
- More informative error messages

Additionally:
- Maybe think about changing target keywords only to: ["mov", "squared"] ?
  Listing all possible channels from our analysis like we do now is probably not very elegant and not helpful for other people that might have other channels (e.g. "squared_accelerometer" or something similar)


timonmerk:
Thanks for refactoring!
I think having `mov_substrs` and `used_substrs` as parameters for the `set_channels_by_bids` function would be cleaner.
This will then be called from `nm_BidsStream` with "mov" and "squared" as deafault arguments

PR #163, opened 2022-02-18
Refactor nm_rereference

richardkoehler:
- Fixed bug with average reference, when channels were marked as "bad"
- Refactored RT_rereference so that reference ids for each channel willnot be re-calculated at each rereference() call, only once when initializing the class instance. This should bring some performance improvement.
- Average references for different channel types (ECOG, LFP, SEEG, LFP) are now automatically handled and not explicitly for each different channel type.
- Had to remove split_data (but I think this wasn't used anyway, right?)

### DIFF
--- a/py_neuromodulation/nm_rereference.py
+++ b/py_neuromodulation/nm_rereference.py
@@ -1,9 +1,10 @@
-from numpy import empty_like, mean, ndarray, where
+"""Re-referencing Module."""
+import numpy as np
 import pandas as pd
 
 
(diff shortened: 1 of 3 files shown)


richardkoehler:
@timonmerk 
Good catch, you are of course absolutely right! Thanks for the fix :)

richardkoehler:
@timonmerk 
Good catch, you are of course absolutely right! Thanks for the fix :)

PR #164, opened 2022-02-19
Add `target_keywords` as argument

richardkoehler:
@timonmerk
As you suggested in the last PR #162, I added `mov_substrs` as an argument to the function:
- I renamed it to `target_keywords`, because you might be decoding something else than movement, so the name is a bit more universal
- As default arguments I added `("mov", "squared", "label")`. I added label here again, so that the default value is a bit more universal and people are not restricted to movement or classification ("squared")
- I changed the default types of used_types and target_keywords to tuples, because probably the default values should be an immutable, right?
- I saw that you changed the default in nm_BidsStream to `used_types:list = ["ecog"]`. So the default in nm_BidsStream now corresponds to `ECOG_ONLY = True`, but before it was `False`. Is that intended? I don't have any preference for the default values, just wanted to double check that this was on purpose :)
- Lastly, I renamed the method from `set_channels_by_bids` to `set_channels`, because I don't think the function actually necessarily uses the BIDS format in its current implementation. But we could think about writing a wrapper in the future, in which you only provide the path to the BIDS `*_channels.tsv` file instead of the arguments `ch_names`, `ch_types` and `bads`.

### DIFF
--- a/examples/example_BIDS.py
+++ b/examples/example_BIDS.py
@@ -21,7 +21,10 @@ def run_example_BIDS():
                                        PATH_BIDS=PATH_BIDS,
                                        PATH_OUT=PATH_OUT,
                                        LIMIT_DATA=False,
-                                       VERBOSE=False)
+                                       VERBOSE=True,
+                                       target_keywords=("SQUARED_ROTATION",),
+                                       used_types=("ecog", "seeg")
+    )
 
     nm_BIDS.run_bids()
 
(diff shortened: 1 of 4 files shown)


timonmerk:
@richardkoehler Thanks again! I fixed the call from `nm_BIDSStream.py` and `nm_stream.py` and adapted the BIDS example

PR #176, opened 2022-05-04
Review nm_resample

richardkoehler:
- Reviewed nm_resample
- Some changes will not be compatible with python 3.9, so I changed the environment file. We can discuss if we want to maintain support for python<3.10!
- Comments that are only meant for the review start with ### -> these need to be deleted before merging into main!
 - Added unit tests for nm_resample


timonmerk (approved):
Many thanks @richardkoehler for the great refactoring and intuitive tests!

I only added a commit shifting some of the environment.yml packages into another env_doc.yml, so a bunch of packages only relevant for running the documentation are left out for the initial environment setup. Next I also fixed a minor bug in the example_BIDS.py. I think this is a pretty nice end-to-end test that comprises features estimation, plotting, and ML analysis. So with the current unit test it also runs through.

Regarding the question about main_rt_test, I agree! This is the example for running real-time estimation (needs major refactoring as well...), but it's definitely not a test, so renaming makes sense.

I also agree shifting the notch_filter.py into nm_filter.py.

PR #177, opened 2022-05-07
Refactor bids stream

timonmerk:
Main change: Generalize Bids Stream to Generic Stream class

I think for many problems data in in BIDS format won't be available. This leaves the main requirements of the ABC nm_stream to be:
 - channel names
 - channel types
 - bad channels
 - rereferencing
 - fs 
 - line_noise
and subsequently defintion of a "data" np array for offline processing.

Thus I renamed `nm_BidsStream` to be `nm_GenericStream`.

Also `scripts/pn_nm_batch_trd_data.py` shows an example of utilizing epoch based data into a "stream" of data. 
This has implications of epoch discontinuities, since concatenating might not be optimal in some cases.

The advantages are though that the general pipeline is followed one to one compared to the "online" pipeline, so all pipeline code, analysis, potting, decoding stays the same.

### DIFF
--- a/py_neuromodulation/nm_GenericStream.py
+++ b/py_neuromodulation/nm_GenericStream.py
@@ -1,4 +1,4 @@
-from typing import Optional, Iterable
+from typing import Optional, Iterable, Union
 import os
 import pathlib
 import mne
@@ -9,11 +9,11 @@ import timeit
 from py_neuromodulation import nm_IO, nm_generator, nm_stream
 
 
-class BidsStream(nm_stream.PNStream):
+class GenericStream(nm_stream.PNStream):
 
     PATH_ANNOTATIONS: str
     PATH_GRIDS: str
-    PATH_BIDS: str
+    PATH_DATA: str
     PATH_RUN: str
     LIMIT_DATA: bool
     LIMIT_LOW: int
@@ -23,10 +23,11 @@ class BidsStream(nm_stream.PNStream):
     raw_arr_data: np.array
     annot: mne.Annotations
     VERBOSE: bool = False
+    READ_BIDS: bool = False
 
     def __init__(
         self,
-        PATH_RUN,
+        PATH_RUN: str = None,
         PATH_SETTINGS: str = os.path.join(
             pathlib.Path(__file__).parent.resolve(), "nm_settings.json"
         ),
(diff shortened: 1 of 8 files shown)


richardkoehler:
@timonmerk  Thanks for the effort, I think actually a class for "generic" data (like numpy or pandas) is super helpful.
But I just had a look at the code to review it and it looked super overwhelming and I was even scared to start reviewing because there are sooo many variables:
grafik
 To be honest, I don't think an __init__ function should look like this, but I think there is maybe a simple solution. The reason that a function has this many variables is usually because it is trying to do many things at once. We are currently overwhelming the "Stream" class with too many concrete tasks or responsibilities. I think the way you conceptualized the "Stream" class is that it should basically only bring the different parts together, i.e. 
1. Fetch data
2. Pass data to "Run" (this returns the features)
3. Pass features to "Projection" (projected features are returned)
4. Pass features to "Model" (this returns a prediction)
5. Save results at the end
But I think the good news is that we have already established the different parts:
1. Fetch data is actually the job of the Stream itself. The information could be passed as the array, DataFrame, BIDSPath, port or other.
2. "Run": Preprocessing (ch_names, references, bads etc.) and feature calculation (fft etc.) are actually defined in the nm_settings.json and nmchannels.csv file standards!
3. "Projection": This information is stored in the grid file!
4. "Model" is loaded from a file (if decoding is wished)
So actually we have already moved all the parameters that we need to define to separate files, that is actually really cool and often recommended. I think it makes the rest much easier.
My ideal vision of using the Stream classes would be the following:
`import neuromodulation as nm

# This could be any type of Stream
stream = nm.XXXStream (
        path_settings = path_settings,
        path_nmchannels = path_nmchannels,
        path_out = path_out,
        path_grids = path_grids,
        verbose = True,
)
stream.load_model(
      path_model = path_model,
)
# This is the step where the stream would initialize the "Features", "Run" etc.
stream.initialize(
       data = np.array(xxxx),
      sfreq = 4000.0
)
# or
# stream.initialize(
#    bids_path = bids_path,
# )
# or
# stream.initialize(
#    port = 1972,
# )
stream.run()
`
What are your thoughts on this and what do you disagree with?

timonmerk:
#178 got merged

PR #178, opened 2022-05-09
Review refactor bids stream

richardkoehler:
Refactor Bids/GenericStream

This is a draft to summarize my ideas concerning the Stream class, related to #177 (the branch is also based on the branch RefactorBidsStream).
Summary:
- All Stream classes are initialized with the same arguments.
- The only abstract method that all Stream classes must share is "run()", so if we want to create new types of streams, we in theory only need to create a new "run()" function.
- The module nm_GenericStream was renamed to nm_stream_offline and now contains the class _OfflineStream and different specific offline Streams: 1. Stream (which is a generic Stream for numpy arrays and pandas Dataframes) and 2. BidsStream (which takes either an absolute file path or an mne_bids.BIDSPath object)
- Offline Streams all inherit from the class _OfflineStream (the underscore because this is not a public class), which in it's turn inherits from the ABC PNStream.
- In the future, we could create in theory create a class _RealTimeStream from which all real time Streams like pylsl-based or FT-based can inherit.
- I had to refactor some functions, but no functionality was really changed. 90% of the changes are in nm_stream_offline and nm_stream_abc
- I updated the example_BIDS and test_example scripts and they run without errors.
- It might be a nice idea to expose the most import functions and classes to the __init__ module of py_neuromodulation. This way you could simply do:
```
bids_stream = nm.BidsStream(...)
```
Let me know what you think :)


timonmerk:
@richardkoehler Thanks again for the refactoring! That decluttered the code and will definitely make it easier to debug issues.

I am not sure though if you actually pushed the most recent version of your local branch? I fixed some stuff that the example_bids ran through without bugs. Some of the main changes were

- leaving out `line_noise` from various classes
- adapting notch filter call in `nm_stream_abc.py` (tbh I am not so sure if it actually makes sense to keep the line_noise param; at least in across cohort analysis becomes necessary..)
- add a `folder_name` and `model `attribute to nm_stream_abc, in this way the `save_after_stream` worked as before
- fixing the `target_keyword `parameter in `example_stream.py` to be a list; otherwise all channels were set to `target `= 1

But I tested the code already in the file scripts/offline_stream_example.py, and it's quite straight forward :)

richardkoehler:
@timonmerk 
Great, thanks for catching those issues! 
Let me briefly comment on your bullet points:
- I agree that `line_noise` doesn't need to be stored everywhere, so that's good!
- I forgot the `model` attribute, and messed up the `target_keyword`, thanks! I pushed an additional fix to allow a single string as `target_keywords` or `used_types`, if you only want to define a single keyword or type.
- Thanks also for fixing the `folder_name` issue. Now thinking about this in more detail, I would probably argue that `folder_name` and `out_path` should not be attributes of the Stream instance. I think they should rather be parameters of the `run` function, because the only place they are used is when the results are saved. So if they are only used in a single place, it doesn't make sense to keep them as an instance attribute. Additionally, for the BidsStream class the `folder_name` parameter wouldn't even be needed, as the `folder_name` is given by the BIDS basename! So `out_path` (or `out_root` to be more precise?) and optionally `folder_name` could simply pass them like this:
```
stream.run(
  ...,
  out_root = xxxx,
  folder_name= xxxx,
)
```
But, I am not 100% sure about this.

- Lastly, I think I need to clarify what I was doing inside the `set_notch_filter` method because I didn't actually explain. So what I noticed is that a lot of the pre-(processing) and feature classes have some parameters, that the user currently doesn't have any access to. For example `NotchFilter` looks like this:
```
class NotchFilter:
    def __init__(
        self,
        sfreq: int | float,
        line_noise: int = 50,
        notch_widths: int | np.ndarray | None = 3,
        trans_bandwidth: int = 15,
    )
```
So there are 4 paramaters: sfreq, line_noise, notch_widths and trans_bandwidth. 
However, the user currently has no way of adapting notch_widths, trans_bandwidth, or the exact frequencies that the notch filter should be applied at. For example, some user might only want to filter the frequencies `[50, 100, 150]`, whereas another user might want to change the `notch_widths`. 
But I thought we could simply use the `nm_settings.json` as an interface to (optionally) set these parameters (the parameters of NotchFilter, but also of other methods, such as BandPassFilter, HighPassFilter, fft etc.). 
The way I imagine it is the user for example sets notch_filter to true in nm_settings.json:
```
"methods": {
        "notch_filter": true,
}
```
Now he has the option to also set the parameters in "notch_filter_settings" (or not, in this case the default values are used). So for example the user could do:
```
"notch_filter_settings": {
    "notch_freqs": [
        50,
        100,
        150
    ],
   "trans_bandwidth": 12
},
```
or not define "notch_filter_settings" at all, in which case an empty dictionary is passed as `kwargs`, and the default arguments are used, which is basically what this line is for:
```
kwargs = settings.setdefault("notch_filter_settings", {})
```
But I didn't want to break backwards compatibility for your code (also, as you mentioned this might be useful for cohort analysis) so I added this line:
```
if "line_noise" not in kwargs:
    kwargs["notch_freqs"] = self.line_noise
```
I think this behavior is also okay, in order to use the same settings file for different cohorts. If `notch_filter` is True, but `notch_freqs` is not defined in the `nm_settings.json`, then the function tries to infer the `notch_freqs` from BIDS file (or maybe for non-BIDS data from a `line_noise` argument in the run function?).
Now we could use the exact same technique with `kwargs` to give the user access to the parameters of all of the methods, like `kalmanFilter`, `fooof`, etc. without having to hardcode these arguments. And we wouldn't have to update the code each time a new argument is added, and backwards compatibility is guaranteed. 
What are your thoughts on this?
As a last but very short note, I wouldn't set 50 as a default value for `line_noise`, rather `None`. Setting 50 as default might lead to some unexpected issues for people working with data from other continents :)

PR #195, opened 2022-10-24
Make pip installable, fix pytest

richardkoehler:
With this PR I propose multiple but rather minor changes (no changes to the way py_neuromodulation works or is structured):

- First I made some minor changes to the tests, in order to make pytest run without error (in testing classes, pytest does not allow an `__init__` method)
- The tests gave a pandas deprecation warning (related to ``df.append()``), which I now circumvent by first appending to a list and only at the end creating the dataframe. This should also be faster and more memory-efficient. The tests still run through smoothly.
- Finally, I moved setup information for pip into the `pyproject.toml` file, which is now the new accepted standard and which most libraries (like black, pylint etc.) also support. I also added some metadata about py_neuromodulation to the `pyproject.toml` file, you can just check it out and add or change anything else.
- You will notice the version of py_neuromodulation is retrieved dynamically from the py_neuromodulation ``__init__.py`` file (this way the ``__version__`` attribute in the py_neuromodulation package and the pip package always match up. I gave it the version number ``0.1.0.dev1`` for the moment, since it is still under development, but feel free to change anything :)
- All requirements for the pip installation are in the ``pyproject.toml`` file, and I added all requirements that were needed for the tests. But I might have missed some other requirements, since the tests currently don't cover all of the modules.
- **Edit:** I noticed black also did some formatting on ``nm_analysis``. But the only relevant changes are those made from lines 895 onward.

@timonmerk I would love to hear your feedback on this. You can try it out by running the following commands:
```
conda create -n pynm-test python=3.10
conda activate pynm-test
```
Navigate to the clone of py_neuromodulation.
```
pip install .[dev]
pytest -v .
```
All tests should optimally run without error.

For editable development mode just do:
```
pip install -e .[dev]
```


timonmerk:
Thanks @richardkoehler, that's super informative and very helpful! You definitely know a lot about the project setup tools :) 
Just one question, by `pip install .[dev]` you're using pip for installation right?

My main impression was so far that conda would be more reliable for that. Do you agree for that? 

Also I definitely need to consult you for documentation after seeing your nice documentation here: https://github.com/richardkoehler/pte-xdf :+1:

richardkoehler:
@timonmerk Yes, ``pip install .`` still only uses pip. But the pip dependency resolver has gotten a lot better in the last 2 years. I have been trying it out in the last few months and I am quite happy with it. I often only use conda for isolating the environment for my project, and then using pip to install all necessary packages. 
You can read here for some more information about the pip dependency resolver if you are interested:
https://pip.pypa.io/en/stable/topics/dependency-resolution/
https://pyfound.blogspot.com/2020/11/pip-20-3-new-resolver.html

Conda probably still is a bit more robust, but it is also more restrictive, in the sense that all dependencies of your package must also be available via conda (I think) for conda to take them into account. And ``py_neuromodulation`` currently has some pip-only dependencies, right? Anyways, I guees if one publishes a package I think the optimal goal would be to make the package available via pip and conda (in the scientific community).

I'm glad that you found pte-xdf and I actually used it to find out how to best build documentation works. We can definitely set up a full documentation for ``py_neuromodulation`` together in a day if you like (once you think we have a fully stable version running)!

PR #223, opened 2023-01-06
Fix tests

richardkoehler:
This PR fixes some minor issues and also fixes some tests that didn't run
- Typo "sess_rigtht"
- Improve handling of float sampling frequency in run_analysis.py
- Update tests to make them run with the recent refactoring
I have to not that the tests that Laura kindly wrote concerning the normalization methods take very long (because the normalization is of course very slow for certain methods), so it is probably not necessary to have these tests run every single time.


timonmerk:
Thanks @richardkoehler! I think the test_normalization test needs to be adapted to only call the normalization function instead of instantiating the stream and running through the whole test examples, which takes way to long. I will work on that

PR #224, opened 2023-01-06
Optimize re-referencing

richardkoehler:
- Re-referencing is now a simple matrix multiplication, which reduces computation time by about 90%
- Added unit test to make sure the new function provides the same results as the old re-referencing


timonmerk:
That's amazing! Thanks a lot. I did not think of rereferencing as a matrix multiplication

PR #281, opened 2024-01-16
Add parallel batch processing

toni-neurosc:
For offline data processing, added the possibility to process batches in parallel processes. This gave a nice 7x speed-up in total run time for my machine (Ryzen 5800H, 16 logical cores) with the following parameters:

NUM_CHANNELS = 5
NUM_DATA = 100000 # Random data points
sfreq = 1000  # Hz
feature_freq = 3  # Hz
n = 3 # Number of runs

Results (saving was disabled for this benchmark):
- Total time sequential: 52.11946733792623 seconds per run
- Total time parallel: 7.207633177439372 seconds per run

I also added a parameter to choose number of threads (although note, they are not technically threads, but actual processes because of how Python works) in case some computing power needs to be reserved on the machine running the code.

Idea: It might be possible to add multi-threading to live data processing too, but at the moment I am not familiar with that part of the program so I haven't determined viability yet.

### DIFF
--- a/py_neuromodulation/nm_stream_offline.py
+++ b/py_neuromodulation/nm_stream_offline.py
@@ -2,6 +2,9 @@
 import math
 import os
 
+import multiprocessing as  mp
+from itertools import count
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


timonmerk:
Thanks also for this pull request. It's definitely something I thought about, and also started to implement myself already at one point.

With the current implementation I unfortunately get a ValueError that I couldn't really figure out how to resolve:

> Traceback (most recent call last):
>   File "c:\code\py_neuromodulation\plot_1_example_BIDS.py", line 106, in <module>
>     features = stream.run(
>   File "c:\code\py_neuromodulation\py_neuromodulation\nm_stream_offline.py", line 327, in run
>     return self._run_offline(
>   File "c:\code\py_neuromodulation\py_neuromodulation\nm_stream_offline.py", line 131, in _run_offline
>     mp.set_start_method(
>   File "C:\Users\ICN_admin\Anaconda3\envs\pynm-test\lib\multiprocessing\context.py", line 251, in set_start_method
>     self._actual_context = self.get_context(method)
>   File "C:\Users\ICN_admin\Anaconda3\envs\pynm-test\lib\multiprocessing\context.py", line 243, in get_context
>     return super().get_context(method)
>   File "C:\Users\ICN_admin\Anaconda3\envs\pynm-test\lib\multiprocessing\context.py", line 193, in get_context
>     raise ValueError('cannot find context for %r' % method) from None
> ValueError: cannot find context for 'fork'

I tried to change it now to joblib, which also seems to work fine. Somehow I had better experiences with joblib in the past and it was also simpler to use. But if you see any disadvantage with that I am also happy to change it back.

toni-neurosc:
Hi Timon, apparently the problem is that an equivalent to the "fork" syscall is not available on Windows, which means that the process creation strategy would need to be changed to "spawn" to make it work on Windows, but that would require using multiprocessing.freeze_support() which requires the main script to have a "if __name__ == "__name__" block which is kind of annoying for the end user. 

Joblib uses its own multiprocessing backend ('loky') that handles all that automatically but adds a bit of overhead, resulting in about 10-15% performance loss in my tests in which is fine imho. Also code is cleaner without try/except:
```Python
        if parallel:
            feature_df = pd.DataFrame(Parallel(n_jobs=num_threads, prefer='processes')(
                delayed(self._process_batch)(batch, n) for batch, n in zip(generator, count(offset_start, sample_add))))
```

Alternatively one could use something like this to choose a strategy based on OS, to maximize performance in all cases:

```Python
match system.platform():
    case 'Linux':
        # use multiprocessing.Pool
    case 'Windows' | 'Darwin':
        # use Joblib
```

toni-neurosc:
Ok, so I pushed the commit with the fix, adding Joblib as a backend, please give it a try on your windows platform in case I made some mistake

Also the way the feature calculation is written currently doesn't really accomodate this section from the original code:

```Python
        if self.model is not None:
            prediction = self.model.predict(features[-1])
```

I don't know exactly what the code is doing and I don't know exactly how to make it fit, perhaps it could be moved inside the new `_process_batch` function?

timonmerk:
Changes merged in #290

Issue #303, opened 2024-04-16
Fix type hints project-wide


toni-neurosc:
Since Pydantic will be used to validate analysis settings, and the contents of settings will be propagated to the different submodules of PyNeuro, type hints should be consistently provided, and correct, all accross the project. MyPy and PyRight are throwing a lot of static analysis errors at the time, which is unclear if they would affect Pydantic integration or not, but I think it's better to fix them, also in case users of the module are using static analysis tools themselves.

[mypy_report.txt](https://github.com/neuromodulation/py_neuromodulation/files/14995650/mypy_report.txt)
[pyright_report.txt](https://github.com/neuromodulation/py_neuromodulation/files/14995651/pyright_report.txt)

timonmerk:
Thanks for running those tests @toni-neurosc.
It would be great to take this as an opportunity to refactor some parts of the current codebase. I had a look at the  files, and found that we can for now ignore:
 - nm_RMAP.py
 - nm_decode.py

For both I need to check what is currently necessary, and how dependencies can be reduced. 

Furthermore, I would even leave out certain modules and delete them in the dev branch for refactoring that would clean up the existing codebase:
  - nm_cohortwrapper.py
  - nm_across_patient_decoding.py
  - nm_eval_timing.py
  - nm_nolds.py (since the computation is way too slow, and not real-time compatible - at least in Python)
  - FieldTrip.py

Apart of that the main point would be to improve the validation with pydantic as an object instead of an dictionary.

toni-neurosc:
Closed with the merge of #305

PR #305, opened 2024-04-23
Fix type hints, clean imports,  and other errors reported by static analysis

toni-neurosc:
So in regards to #303 I passed 3 different static analyzers over the codebase (Ruff, PyRight and Mypy) and fixed a bunch of stuff of varying grades of relevance. The goals were:
- To overall improve the consistency of the code and adherence to latest standards
- To make sure default values for arguments are handled correctly (e.g. in the case of None defaults)
- To clean up error reports of trivial warnings and flags so that potential serious errors are easier to spot in future static analyses
- To perhaps improve startup time by cleaning unused imports

In terms of results, they were:
- For Mypy, going from 272 errors to 24 errors. These 24 errors are now things that actually require fixing, see report file below generated by this command:
```mypy --follow-imports skip --exclude nm_cohortwrapper\.py --exclude nm_across_patient_decoding\.py --exclude nm_eval_timing\.py --exclude nm_nolds\.py  --exclude FieldTrip\.py --exclude nm_RMAP\.py --exclude nm_decode\.py  py_neuromodulation\```

[mypy_report_23_04_24.txt](https://github.com/neuromodulation/py_neuromodulation/files/15074254/mypy_report_23_04_24.txt)

- For PyRight, from [287 errors, 2 warnings] to [168 errors, 0 warnings]
[pyright_types.txt](https://github.com/neuromodulation/py_neuromodulation/files/15075289/pyright_types.txt)

A non-exhaustive but quite complete list of the changes done are:
- Remove unused imports
- Fix incorrect type typehints (e.g. bool = None, or np.array instead of np.ndarray)
- Change Optional and Union typehints to modern style i.e. list[str|None] or int | None
- Add comments to potential bugs, check commit (e.g. duplicated "s" argument to axis.scatter)
- Fix empty error raises with corresponding errors
- Commented out variables that were created but never used, feel free to remove
- Commented out class-level variables that were intended to be instance-level variables (i.e. variables defined at top-indentation level within a class and without self. prefix)
- And a bit of perhaps a controversial one: Favor empty lists and empty strings as default values for arguments as opposed to None values. Reasons are several: 
    - static analysis complains a lot about None obviously not having any methods or being passable to functions. 
    - it generates a lot of type analysis cascades of errors (e.g. when the input to a function can be str | None but that input is passed to a function that only accepts str and doesn't even handle the None case)
    - If i'm not mistaken, passing empty lists, arrays or strings is never an expected input for the functions where I changed this, so handling the empty case instead of the None case should not generate any unexpected behaviours

I would understand if the last change is not accepted, for me it seems more intuitive to check for empty inputs or sentinel values rather than Python NoneType, but it's a matter of preference at the end of the day.

### CODE REVIEW

File: py_neuromodulation/ConnectivityDecoding/_get_grid_whole_brain.py [hunk 1]
```diff
@@ -1,18 +1,15 @@
...
-
 class NiiToMNI:
 
     def __init__(
         self,
         PATH_nii_file: str = r"C:\code\RMap_ROI_Estimation\Automated Anatomical Labeling 3 (Rolls 2020).nii",
     ) -> None:
 
-        self.img = nib.load(PATH_nii_file)
+        self.img = nib.Nifti1Image.from_filename(PATH_nii_file)
```
timonmerk, line 12: I think those files need to be restructured in further detail by me. I would for the mypy validation leave out the whole ConnectivityDecoding folder for now.
But thanks for addressing those also already.

py_neuromodulation/ConnectivityDecoding/_get_grid_whole_brain.py
```diff
@@ -74,22 +71,22 @@ def select_non_zero_voxels(
...
-    def plot_3d_coordinates(self, coord_non_zero: np.array):
+    def plot_3d_coordinates(self, coord_non_zero: np.ndarray):
         fig = plt.figure()
         ax = fig.add_subplot(111, projection="3d")
         ax.scatter(
             coord_non_zero[:, 0],
             coord_non_zero[:, 1],
             coord_non_zero[:, 2],
-            s=50,
+            s=50, # Bug? Third and fourth argument are s
```
timonmerk, line 89: That's a 3D plot, so `s` is just the scatter point s.

py_neuromodulation/FieldTrip.py
timonmerk, line 1: This is a file taken externally from here: https://github.com/fieldtrip/fieldtrip/blob/master/realtime/src/buffer/python/FieldTrip.py, I would therefore not change the syntax in this file and only mention maybe as a comment that's it's an external file with the given link.

py_neuromodulation/__init__.py
timonmerk, line 1: Thank's for catching the wrong `__version__` tag.  I've figured out that's wrong. Probably there is a way to dynamically take that from the `pyproject.toml`? I guess that would be quite useful and it's also done in some other packages to quickly retrieve the version.

py_neuromodulation/nm_RMAP.py
```diff
@@ -351,8 +351,8 @@ def get_fingerprints_from_path_with_cond(
 
     @staticmethod
     def save_Nii(
-        fp: np.array,
-        affine: np.array,
+        fp: np.ndarray,
```
timonmerk, line 354: Interesting, just figured out that `np.array` is a function and `np.ndarray` is a class. So thanks for catching! :)

py_neuromodulation/nm_analysis.py
```diff
@@ -23,21 +22,22 @@
...
-    sidecar: dict
-    sfreq: int
-    line_noise: int
-    nm_channels: pd.DataFrame
-    feature_arr: pd.DataFrame
-    ch_names: list[str]
-    ch_names_ECOG: list[str]
-    decoder: nm_decode.Decoder = None
+    
+#     These are class variables, not instance variables
```
timonmerk, line 26: That's great. Makes total sense. Back then I was not aware of the difference.. 
Could you delete the comments? Thanks!

py_neuromodulation/nm_analysis.py
```diff
@@ -210,7 +211,7 @@ def filter_features(
         return feature_select
 
     def set_target_ch(self, ch_name: str) -> None:
-        self.label = ch_name
+        self.label_name = ch_name # Toni: I think this was the intended behavior
```
timonmerk, line 214: Yes! Great catch

py_neuromodulation/nm_analysis.py
```diff
@@ -921,31 +922,28 @@ def set_score(
...
                     # transform dict into keys for json saving
 
                     set_score(
                         key_set="mrmr_select",
-                        key_get=None,
                         take_mean=False,
                         val=obj_read["mrmr_select"],
                     )
-                if model_save is True:
+                if model_save:
```
timonmerk, line 944: Did mypy complain about those checks? Somehow those checks were for me more interpretable that I directly know `if x is True` then x needs to be boolean. Otherwise x could be any datatype..

py_neuromodulation/nm_cohortwrapper.py
```diff
@@ -301,29 +286,30 @@ def read_all_channels(
...
-            sess_name = feature_file[
-                feature_file.find("ses-") + 4 : feature_file.find("_task")
-            ]
-            task_name = feature_file[
-                feature_file.find("task-") + 5 : feature_file.find("_run")
-            ]
-            run_number = feature_file[
-                feature_file.find("run-") + 4 : feature_file.find("_ieeg")
-            ]
+            # Unused variables, remove?
```
timonmerk, line 294: Yes, if they are not used I would remove them

py_neuromodulation/nm_decode.py
```diff
@@ -39,56 +33,58 @@ def __init__(
...
-    rus: RandomUnderSampler = None
-    VERBOSE: bool = False
-    ch_ind_data: dict = {}
-    grid_point_ind_data: dict = {}
-    active_gridpoints: list = []
-    feature_names: list = []
-    ch_ind_results: dict = {}
-    gridpoint_ind_results: dict = {}
-    all_ch_results: dict = {}
+    # TONI: this are defined as class variables, not instance variables
```
timonmerk, line 63: Yes, thanks! It would be great to remove the comments

timonmerk: But it's very helpful to see it that way!

py_neuromodulation/nm_filter.py
```diff
@@ -1,13 +1,12 @@
...
-
-logger = logging.getLogger("PynmLogger")
-
 import mne
 from mne.filter import _overlap_add_filter
 import numpy as np
+from typing import cast
 
+import logging
+logger = logging.getLogger("PynmLogger")
```
timonmerk, line 9: I guess those are the same changes as in the logger PR?

py_neuromodulation/nm_generator.py
```diff
@@ -6,7 +6,7 @@
 def raw_data_generator(
     data: np.ndarray,
     settings: dict,
-    sfreq: int,
+    sfreq: int | float, # Added float, consistent with rest of sfreq variables
```
timonmerk: Yes, great!

py_neuromodulation/nm_linelength.py
```diff
@@ -1,18 +1,18 @@
...
+class LineLength(Feature):
     def __init__(
         self, settings: dict, ch_names: Iterable[str], sfreq: float
     ) -> None:
         self.s = settings
         self.ch_names = ch_names
 
     @staticmethod
-    def get_line_length(x: np.ndarray) -> np.ndarray:
+    def get_line_length(x: np.ndarray) -> np.floating: # np.mean w/o axis argument
```
timonmerk, line 15: That's the mean across all axes

py_neuromodulation/nm_plots.py
```diff
@@ -54,17 +53,18 @@ def plot_df_subjects(
...
         handles, labels = ax.get_legend_handles_labels()
-        l = plt.legend(
-            handles[0:n_hues],
-            labels[0:n_hues],
-            bbox_to_anchor=(1.05, 1),
-            loc=2,
-            title=hue,
-            borderaxespad=0.0,
-        )
+        # Never used, remove?
```
timonmerk, line 59: The legend is used, but not the `l = ` expression. So `l` can be removed

py_neuromodulation/nm_sharpwaves.py
```diff
@@ -256,8 +263,8 @@ def analyze_waveform(self) -> None:
 
         """ Find left and right peak indexes for each trough """
         peak_pointer = 0
-        peak_idx_left = []
-        peak_idx_right = []
+        peak_idx_left_list = []
```
timonmerk, line 266: That's great! More intuitive to know it's a list

py_neuromodulation/nm_settings.py
```diff
@@ -3,15 +3,15 @@
 
 import pandas as pd
 
-from py_neuromodulation import nm_IO
+from py_neuromodulation.nm_IO import PYNM_DIR, read_settings
```
timonmerk, line 6: Do you think it could make sense to add PYNM_DIR to the `__init__.py` file? Intuitively I would have guessed it would belong there

timonmerk (changes requested):
Thanks a lot for all that changes! That was tremendous amount of work! I left a couple of comments (and questions 😄) but it's mainly minor things.

The fieltrip file I would not change. And a couple of other files would kind of need to "refactored-out". It's great that you went through them now to adapt the mypy output but I would mainly focus, also wrt the pydantic setup, on the preprocessing and feature modules.

I would have a look at the mypy errors and commit them also to this branch later this week.

py_neuromodulation/nm_settings.py
toni-neurosc: Probably yes. It was mostly used in nm_IO so I defined it first there, then I found it used in other places. I already moved it there.

py_neuromodulation/nm_sharpwaves.py
toni-neurosc: This was mostly to stop MyPy from complaining about redefining variables, but since I'm used to C++ where you just can't never modify type for me it also makes sense, it helps keep tracking of the manipulations. Also if that JIT compiler thing with Python ends up being relevant (or in the event we can make PyPy work with the project) having distinct variables by type proably helps too.

py_neuromodulation/nm_linelength.py
toni-neurosc: Sure, I was just clarifying why I changed the return type.

py_neuromodulation/nm_generator.py
toni-neurosc: Actually I wanted to ask about this. Sfreq is in many places allowed to be both `int` or `float`, but then in other places it is allowed only as `int` or coerced to `int`.  So, at the time of feature calculation, is the sampling frequency always supposed to end up as a whole number? Because in that case we can handle it at the settings validation level.

py_neuromodulation/__init__.py
toni-neurosc: So I was looking into that yesterday and there are ways to make it dynamic in both directions: either have pyproject.toml be dynamic or have `py_neuromodulation.__version__` be dynamically assigned from the metadata.

But the most common answer to the question that I found is that the `__version__` variable was never a Python standard and that nowadays is kind of useless. So I removed it altogether in favor of the pyproject metadata, but we can always have it back.

py_neuromodulation/FieldTrip.py
toni-neurosc: OK, will undo the changes and do as you say.

py_neuromodulation/ConnectivityDecoding/_get_grid_whole_brain.py
toni-neurosc: Interesting, in the method definition it looked like it did not accept a third axis and Mypy was complaining, it goes like `def scatter(self, x, y, s=None`. The 3d axes class is actually defined in `mpl_toolkits.mplot3d.Axes3D`, the `projection='3d'` thing must be returning that class instead of the normal axis. Will leave it be then.

py_neuromodulation/nm_analysis.py
toni-neurosc: Sure.

py_neuromodulation/nm_analysis.py
toni-neurosc: Oh, I see what was the point of it. Well, this is probably one of those issues where personal preference and coding style makes a huge difference. But I don't think we should waste a lot of time on it so I'll explain my view and you make the call:

- In terms of Python style guide, indeed both the official documentation and the PEP8 style guide recommend against comparisons with True:
    - https://docs.python.org/2/library/stdtypes.html#truth-value-testing
    - https://peps.python.org/pep-0008/
- Is correct that `is True` checks will fail if the variable passed is not a boolean, but we have other levels to account for that:
    - We have type hinting all around the project
    - We have type checking implemented statically and in the close future also dynamically with Pydantic
    - Variable names usually should be enough to tell if a value is a boolean or not. This again comes down to style and personal preference, but when it comes to naming variables changing `model_save` for `save_model` or `saving_enabled` goes a long way.

On the other hand:
- It's pretty much harmless and if it helps you navigate your own project, that's a good reason to keep it. The `is True` part can always be replaced with a single Regex project-wide replacement (what I did basically xD)

Sorry for the long text, it's really not that big a deal xD I'm just that much of a nerd about this stuff i guess.

py_neuromodulation/nm_filter.py
toni-neurosc: Not exactly, this was just moving the logger initialization below all the other imports, so that linters would not complain about imports not being on top xD Just an aesthetic thing

py_neuromodulation/nm_plots.py
toni-neurosc: Oh, true, what was i thinking there... been a long time since using pyplot sorry

py_neuromodulation/nm_analysis.py
timonmerk: Thanks for the thorough explanation! Pep-0008 actually has a section about boolean comparison:

> Don’t compare boolean values to True or False using ==:
> # Correct:
> if greeting:
> # Wrong:
> if greeting == True:
> Worse:
> 
> # Wrong:
> if greeting is True:

So they're basically arguing against the `is True`. But I think it's a minor point, and I am fine either way, especially if we have type hints it's not a big point.

py_neuromodulation/__init__.py
timonmerk: I found [this](https://stackoverflow.com/a/67097076/5060208) solution to add the __version__ property as a separate commit now

py_neuromodulation/nm_generator.py
timonmerk: Thanks for addressing that point, in general it would be great to always the same sfreq value being int or float. So it's great that you added here the float type.
During data retrieval in the nm_generator the float type needs to be casted then into int for correct indexing.

timonmerk:
@toni-neurosc  thanks again for those changes and mentioning about the mypy errors. I tried to solve them and also fixed a couple of them.

It raised an issue with the preprocessing classes. Currently there is no good method to call the tests. I thought about it for some time now but either the tests are called locally within the init method, which is a good way to ensure also testing before instantiation.
But on the other hand it would be nice to call it also externally from nm_run_analysis when the preprocessors are instantiated..

Plus there is not enough testing implemented for all preprocessors that sometimes the functions would be empty. So I decided to remove the testing method currently. Not sure, we could also have a similar implementation with the features that they are called on the class with the same params but then also some tests brake s.t. the instantiation without nm_run_analysis should also break with wrong params, meaning that I want to call the tests from the init method.
Quite a design question.. but I would say let's leave it for now with this method even if it's not super elegant.

py_neuromodulation/nm_generator.py
toni-neurosc: OK, so this has been resolved in one of the new commits I added that basically changes all types that were "int|float" to just "float", because actually for the type hinting standard they are exactly the same. Float implies Int basically.

Also, Python does not really know about float or int types like other languages do, it just interprets the numbers as it goes so there is no danger of mixing low level numeric representations.

So after that, now sfreq is basically passed around as a real number until it hits the DataProcessor constructor, at which point it bifurcates into `sfreq_raw` (which is truncated by integer division) and `_sfreq_raw_orig` which is kept as a copy of the original value. Does that check out?

toni-neurosc:
Hi @timonmerk, don't worry about the test_settings methods, that will be lumpted together with the settings class, the FeatureSettings class/protocol, and the preprocessor settings class/protocol. In case a preprocessor does not have settings to test, i don't think its a problem to leave the function empty, but they way you left it now looks good to me for the time being.

Also, I did some minor changes to the testing scripts, just removing unused imports and fix a couple types that didn't match. I think it's not affecting function but since you know better than me the purpose of each test is best you take a look at it.

Also, I realize I messed up the type and default value of the "clip" parameter for _normalization_and_clip, now it's fine, it defaults to 0 which when arrives to the normalization function will test `False` and be replaced with the default 3.

toni-neurosc:
I feel like there are some changes that I have not explained and I should mention here for clarity:

- Path manipulation was being done with `os` and `pathlib` interchangeably. Since `pathlib` is the more "modern" Python approach  and it's being actively developed I decided to change all the instances of `os` to the `pathlib` equivalents, and also simplified some string manipulation logic in `nm_IO.py`.
    - The only exception where I could not get rid of `os` was one instance of `os.walk()`, that in Python 3.10 does not have an equivalent in `pathlib`. In Python 3.12 `Path.walk()` was introduced so when we bump the Python version requirements to 3.12 or higher we can uncomment that line.
- I removed some dependencies in pyproject.toml that were not being used in the project. They were:
    -  `pybids`
    - `pynput`
    - `pip`
    - `wget` -> This one because it's not maintained and `urllib.requests` can replace it easily
- I added `statsmodels` as a dependency, it was being used but not listed, probably it was itself a dependency of one of the packages I removed.
- I realized I removed comparisons to True but not to false so I have replaced instances of `if ____ is not False:` with `if not ____:`

timonmerk:
@toni-neurosc Thanks a lot for all those changes. It improved the readability of the code quite a lot and will hopefully also make the integration with pydantic easier.
Unfortunately it is now at a stage where it's very difficult to review the changes, and it's many files that have been changed. So I propose that you make the merge into main yourself. Unless there are specific files that you think I need to check?

I tried to merge the pyproject.toml but there was a syntax error. Additionally, I merged the GitHub workflow PR already into main, so it may require checking that those changes are compatible.

I think those changes do not brake backwards compatibility, since it doesn't break the API. But once the tests run through, now also for Windows, feel free to merge into main.

toni-neurosc:
Thanks a lot for trusting me with the merge @timonmerk ! I have been working today on this for quite a while and I think it's safe to merge, I did indeed introduce a couple bugs that I had to solve, but I think they were the only ones and since they were related to typing and input validation anyway, if there are other bugs that I missed they will most likely be caught soon enough in my next PRs. So I'm closing this one up, and I hope I never have to submit such a massive PR again xD

PR #306, opened 2024-04-24
Add logging class 

toni-neurosc:
So I liked a lot the idea of using the logging library to replace prints, but the way it was implemented it was not taking full advantage of the logging system so I created a new module called nm_logger with a logger class inheriting from the original logging.Logger class with some tunable defaults and extra methods. 

This way we can choose the defaults that are more reasonable for this software and also we can extend the functionality of our logging system at will modifying only this class. 

Current defaults are log to console at "INFO" level. I also put a function to log to a file there as a demo of the things we can implement.

I got the general idea on how to structure the system from here, and created the logging class on top of it: https://dev.to/joaomcteixeira/setting-up-python-logging-for-a-library-app-6ml


timonmerk (changes requested):
That's great! Thank you for addressing the logger. I think the changes make the logging more clean.

I have a feature request however, that the previous logger did not include.

When running the logger it would be great to save also the `logfile_pynm.log` in the stream output directory in addition with the features.csv etc.

In addition, the timing console prints currently write out timing precise to 2 decimal points, but the output log files write out mostly one decimal point:

> 2024-04-15 16:24:47,447:INFO:PynmLogger:nm_run_analysis.py:Last batch took: 0.04 seconds
> 2024-04-15 16:24:47,452:INFO:PynmLogger:nm_run_analysis.py:Last batch took: 0.0 seconds
> 2024-04-15 16:24:47,457:INFO:PynmLogger:nm_run_analysis.py:Last batch took: 0.0 seconds
> 2024-04-15 16:24:47,461:INFO:PynmLogger:nm_run_analysis.py:Last batch took: 0.0 seconds

It would be great to keep therefore the two decimal points also in the saved log output file.

Many thanks!

toni-neurosc:
Hi Timon, glad you liked the change. Regarding the logging to a file, there's different ways to do it: 
- have it so that the same output that goes to the console goes ALSO to a file. 
- have it the above way by default but also adding the possibility of changing the level of the console and file logging independently
- the way i have it in the log_to_file function, that saves 2 separate files: one for INFO level and one for DEBUG level. Maybe this could be turned into a function that enables a pynm_debug.log on top of the regular one? Like e.g. `logger.set_debug_logfile(state: bool)` or maybe a ` @property` attribute

Also probably the logfile name should include datetime data too.

Will look into the formatting thing

timonmerk:
@toni-neurosc  Thanks again for improving the logger. 
I added the minor call to your log_to_file function which worked perfectly out of the box. I think having the call in the run function is a good place since it's the first time when output paths get specified. 
Feel free to merge if you agree with the changes.

toni-neurosc:
Sure, having the logs in the data output folder is a good idea. I changed the directory creation to Path.mkdir (a bit less verbose) and fixed the decimal point issue. It was a matter of using the Python f-string formatting function instead of np.round, like this: 
```python
f"Last batch took {time():.2f} seconds"
```
It uses the same format as C's printf so I'm pretty used to it by now xD

I'm merging it now as it is, the logger can get more features in the future but I think for now we've done enough.

PR #308, opened 2024-04-26
Move tox.ini contents to pyproject.toml

toni-neurosc:
What says in the title, I created the PR to see if the tests run fine


toni-neurosc:
Ok so that works 👍

toni-neurosc:
Ok so i'll just merge this myself since it's a pretty harmless change

Issue #310, opened 2024-04-26
Use Rye to install dependencies during automated tests


toni-neurosc:
So @timonmerk and I talked about this during the meeting, and it seems that there is a way to customize the way tox installs packages, see docs here: https://tox.wiki/en/latest/config.html#install_command

So it defaults to `python -I -m pip install  {opts} {packages}`

Right now the output in the Github Action is:
```python -I -m pip install 'filterpy>=1.4.5' fooof 'hatch>=1.9.4' imbalanced-learn 'joblib>=1.3.2' 'matplotlib>=3.3.4' mne 'mne-bids>=0.8' mne-connectivity mrmr-selection nolds 'numpy>=1.21.2' 'pandas>=1.2.2' pip 'pyarrow>=14.0.2' pybids 'pybispectra>=1.0.0' pynput pyparrm scikit-image 'scikit-learn>=0.24.2' scikit-optimize 'scipy>=1.7.1' 'seaborn>=0.11' wget```

So we can experiment with adding this to the tox configuration:
```  
[testenv]
install_command = rye sync
```
Provided that the Github workflow is modified to install Rye beforehand, instead of Pip.

We could maybe do this modification together with adding the Windows tests, @SamedVossberg

toni-neurosc:
Done, PR #311 fixes this

PR #311, opened 2024-04-27
Improve Github actions testing workflow

toni-neurosc:
So, as I mentioned in issue #310 I wanted to see if Rye could speed up setting up the testing environment for Github actions automated testing.

In the end I made several changes:
- Rye was not the adequate tool, is more a development environment manager than a package installer, so I used uv (https://astral.sh/blog/uv) instead, which is created by the same developers and it's what Rye uses under the hood.
- The Github action for setting up uv (https://github.com/drivendataorg/setup-python-uv-action/) also handles dependency caching which is a cool feature, although even without the cache the whole process if very fast (it's about 7 seconds to download and install everything)
- I got rid of Tox since it wasn't doing really anything for us at the moment. Right now the only real work it was doing was installing pytest, and that can be easily handled in the Github action workflow.
- I added pytest-xdist which multithreads the tests, basically running 4 cores at the same time and using them to run tests in parallel.

Overall, the testing time in Linux went from 2m20s to 1m10s, so that's half the time if my math is correct.

PS. Oh, and forgot to add that tests run in Windows now as well

PS2. With dependencies already cached, 59s on Linux!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,34 +20,36 @@ on:
       - '*.txt'
 jobs:
   tests:
-    name: ${{ matrix.name }}
-    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.platform.name }} Python ${{ matrix.python }}
+    runs-on: ${{ matrix.platform.os }}
     env:
       APPDATA: DefaultNonExistingPath 
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - {name: 3.10 Mac, python: '3.10', os: macos-latest, tox: py310}
-          - {name: 3.10 Linux, python: '3.10', os: ubuntu-latest, tox: py310}
+        platform:
+          - os: ubuntu-latest
+            name: Linux
+            activate_env: . .venv/bin/activate
+          - os: macos-latest
+            name: MacOS
+            activate_env: . .venv/bin/activate
+          - os: windows-latest
+            name: Windows
+            activate_env: .venv\Scripts\activate
+        python:
+          - '3.10'
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@v4
+      - name: Set up Python with uv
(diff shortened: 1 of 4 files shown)


timonmerk:
Wow, that's amazing @toni-neurosc! Thanks soo much looking into that and fixing this in such an elegant way!

Do you think it makes sense to also recommend to use the uv installer as recommended tool?
Currently I mentioned in the ReadMe rye, but I think it still uses pip under the hood.

I would be curious to know how you would recommend that. I saw in our meeting you used conda to create an environment, but for my machine the dependency resolving was with conda way too slow so I tried to recommend rye and leave out conda altogether..

toni-neurosc:
Hi @timonmerk . So, Rye does indeed use 'uv' as an installer, it's all developed by the same people (Ruff, Rye, uv).

The problem I see with Rye is not that is not a good tool, it's that it serves a different purpose. Rye is more like, a dev environment manager, and a build/deploy tool. So in that sense, it's more like an alternative to Hatch, and in the "contributing guidelines" that you mentioned in another message, we should tell people to use either one or the other, and choose one to manage our dev tools and dev environment, building docs, publishing and all that fuss.

In fact using Rye might be problematic for end-users, since they need to sync their environment with our project with `rye sync`, conform to our version Python, and activate the .venv created by Rye which they might not want to do e.g. if they're developing their own tool that depends on PyNM and they just want to add it as a dependency, in which case they should be doing `rye add py-neuromodulation` and not `rye sync`!

On the other hand 'uv' is more like a faster version of pip, in fact is almost a drop-in replacement, so you could just in theory do `alias pip="uv pip"` and it would work pretty much the same but faster.

So I think in our "install guide" (not for devs, just for users) we should recommend doing either `pip install py-neuromodulation` or `uv pip install py-neuromodulation`. They'll get the package the same, but we can throw 'uv' some support by recommending that people use it 😄

timonmerk (approved):
Thanks a lot @toni-neurosc for this super useful PR. It really makes sense trying to improve the speed of those automatic tests since they run basically every time code changes are made!

PR #317, opened 2024-04-28
Add NMPreprocessor protocol and Preprocessors class 

toni-neurosc:
This change gives Preprocessor classes the same treatment that Feature classes were being given in my other PR #307 , that is:

- I created a Protocol "NMPreprocessor" from which all preprocessor classes inherit
- I moved the initialization and execution of preprocessing from "nm_run_analysis" module to its own module "nm_preprocessing"
- Like the class Features, I created a class Preprocessors that imports the necessary modules according to settings and instantiates them.
- The NMPreprocessors instances are then run on the data by using `Preprocessors.process_data(data)`

I also added some utilities that can be shared by nm_features too:
- a NamedTuple class in nm_types called `ImportDetails` (not sure I like the name a lot) that holds data about the name of a class and the module it's located in
- a function get_class that lazy imports a class and returns the class object, given an `ImportDetails` input. Right now I put it in nm_types, but i don't know if that's the right place for it

This was quite a difficult modification due to the differences in the arguments taken by the different Preprocessor classes, so it requires quite a bit of "meta-programming" to get it working without errors. But I took some valuable lessons regarding typing and also regarding the different "settings" needs of the different preprocessors.


timonmerk (approved):
Great changes @toni-neurosc! Also very nice solution that it's a similar modular import of the preprocessors omitting the switch case.

Btw the test fails due to a bug in the test file where the sampling rate was parametrized wrong, it was already fixed in another PR

toni-neurosc:
Awesome, merging #305 here did indeed fix the testing errors. Merging this into main then.

PR #318, opened 2024-05-01
Merge main into LSL branch

toni-neurosc:
I reviewed the merge conflicts between the refactored main branch and the LSL branch, there were actually not so many of them, check the changes and merge please.

EDIT:
So I have made a bunch more changes needed to fix the tests, some of them are temporary and need to be addressed:
- Added xvfb (X-server virtual framebuffer) to run the tests on linux, since `pynput` depends on Xorg and since the Github runner is headless it was throwing an error. It's a fix for now but we need to find a solution for this... (or maybe its fine like this?)
- I had to change `raw_data_generator` to return a tuple of `[None, data]`, same as LSLStream get_next_batch(), otherwise tests would crash on wrong number of outputs for unpacking
- I replaced downloading/installing `liblsl` with including the dynamic library files in the repo itself, under the "liblsl" folder, and using the Github actions matrix to set the proper `MNE_LSL_LIB` variable for each OS. We need to figure out how to deal with shipping this libraries when building with Hatch to ship to PyPI
- Changed Pyside6 dependency to pyqt5 cause the former was giving me errors.

PS. The Qt dependency is introduced by mne_lsl class StreamViewer. Are you using this class only for testing purposes @SamedVossberg? I think we should not rely on third party GUI apps/widgets when we start developing our own GUI, otherwise we will be forced to use the same backend/library as the tools we depend on (in this case Qt5)


toni-neurosc:
So at some point I had it working... but now it for some reason it's crashing, and the crash dumps suggest it has to do with the Xorg server, but XVFB was working fine at some point, not sure why it suddently started crashing the Github runners...

toni-neurosc:
Alright, problem solved, I changed the matplotlib backend to Tkinter in __init.py__
```Python
import matplotlib

# Set matplotlib backend to TkAgg because it crashes with the Qt5 install we're using for LSL
matplotlib.use("tkagg")
```

Well that was a real pain in the ass to figure out... but good to know for the future

py_neuromodulation/nm_generator.py
```diff
@@ -143,12 +141,7 @@ def raw_data_generator(
...
-
         if (cnt - offset_start) >= ratio_samples_features * ratio_counter:
-
             ratio_counter += 1
 
-            yield (
-                np.arange(cnt - offset_start, cnt) / sfreq,  # what is this exactly? Not in main. need? 
-                data[:, np.floor(cnt - offset_start).astype(int) : cnt]
-            )
+            yield None, data[:, np.floor(cnt - offset_start).astype(int) : cnt]
```
SamedVossberg, line 147: We might want to change this back to include the timing/counter.
I'll do this.

SamedVossberg (approved):

PR #325, opened 2024-05-08
Minor changes on top of mne_lsl merge

toni-neurosc:
Hi @timonmerk, I made some changes on top of the recent LSL merge. Summary:

- Remove settings parameter from LSLOfflinePlayer since it's not used, as the sampling frequency  is passed in another parameter
- Defer LSLStream imports
- Fix a typing error
- Remove a couple variables
- Changed a couple errors for warnings:
    - If both a filename and a data array was passed to LSLOfflinePlayer, it would throw an error. I changed it to a warning and defaulted to using the file as data source.
    - If _GenericStream was reading from an LSLStream with different frequency that the one that was passed it would  throw an Exception, instead I think we can warn the user and default to the LSLStream frequency.

It's in general problematic to have 2 sources of truth like in the case of GenericStream getting an sfreq value from the user and one from the stream. It's also quite annoying that Python does not support overloaded constructors like C++. Maybe when I'm done with the Settings I will also incorporate Pydantic into the stream classes to handle this better.

### DIFF
--- a/py_neuromodulation/nm_generator.py
+++ b/py_neuromodulation/nm_generator.py
@@ -4,44 +4,43 @@ import time
 from pynput import keyboard
 import numpy as np
 
-from mne_lsl.player import PlayerLSL
-from mne_lsl.stream import StreamLSL
-import mne
-
 from py_neuromodulation import logger
 
 class LSLOfflinePlayer:
 
     def __init__(
         self,
-        settings: dict,
         stream_name: str = "example_stream",
         f_name: str | None = None,
         sfreq: int | float | None = None,
         data: np.ndarray | None = None,
     ) -> None:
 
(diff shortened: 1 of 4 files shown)


timonmerk:
Thanks @toni-neurosc for that PR. Since there were a couple of changes in the branch refactor_lsl_stream (https://github.com/neuromodulation/py_neuromodulation/tree/refactor_lsl_stream) I would not directly merge it.
Maybe you can check @SamedVossberg to include those lazy import changes in your next PR

toni-neurosc:
ok @timonmerk I have change this PR then to be a merge into the `refactor_lsl_stream` branch, at some point I will resolve the conflicts and let @SamedVossberg merge if he's alright with that

PR #330, opened 2024-06-07
nm_train live training class

SamedVossberg:
- Added the nm_train file which opens a Pyside6 GUI to record training data and train classifier
- Added a disconnect method the lsl stream
- Added a jupyter notebook example for the Trainer class


timonmerk:
wish list:
 - write `nm_db.py` module which creates sqlite tabel in outpath if it doesn't exists 
 - add sqlite write function `nm_db.write_features` after feature computation in run method
 - remove list aggregation
 - add function in `nm_db.py` to write head of db as csv and option to read db as pd df
 - return value of `run` method is pd dataframe of database
 - replace `nm_analysis` read db as df instead csv file

-> read into option to load sqlite db in webframe

timonmerk:
Closed now and adressed in issue #345

PR #337, opened 2024-06-17
Timing lsl

SamedVossberg:
- changed times and logging
- excluding parallel processing


timonmerk:
Great! Many thanks @SamedVossberg!

PR #339, opened 2024-06-17
MNEFilter with fftconvolve

toni-neurosc:
Simple change, just change `np.convolve` to `scipy.fftconvolve` to do the convolution on all channels at once. Reduces `nm_bursts` computation time by 14% in my benchmark.

Output is the same, save floating point error (identical up to the 13th or 14th decimal place).

Had this in the backburner waiting for the #329 merge, didn't want to add more non-settings changes to that PR.


timonmerk:
Many thanks @toni-neurosc! Makes a lot of sense to add it, also since it's present now in the sharpwave analysis.

PR #341, opened 2024-06-18
Add compatibility with Numpy 1.26 and 2.0

toni-neurosc:
Hi @timonmerk, I saw that you bumped the package requirements to Numpy 2.0.0 and Pandas 2.2.2, which is fine I think, but maybe it could cause conflicts for end-users. Like, I already found that StreamLit does not support Numpy 2.0.0 yet, so it rolled back to 1.26 when I installed it. 

So I added code to be compatible with both old and new versions of Python, all tests are passing and pip is installing the requirements just fine.

I changed the requirements back to Numpy 1.26 (for Python 3.12 compatibility) and Pandas 2.0.0 just to pick a round number.


timonmerk (approved):
Makes sense! Thanks!

PR #349, opened 2024-06-24
Remove features dictionary as input to calc_feature

toni-neurosc:
The calculation of each feature is independent of the rest, so having "features_compute" as an input argument to `calc_feature` suggested otherwise and was not necessary. It might also pose problems if we decide to multi-thread the feature calculation at the `DataProcessor` or `FeatureProcessors` level.

### DIFF
--- a/examples/plot_2_example_add_feature.py
+++ b/examples/plot_2_example_add_feature.py
@@ -37,18 +37,21 @@ class ChannelMean:
         
         self.feature_name = "channel_mean"
 
-    def calc_feature(self, data: np.ndarray, features_compute: dict) -> dict:
+    def calc_feature(self, data: np.ndarray) -> dict:
+        # First, create an empty dictionary to store the calculated features
+        feature_results = {}
+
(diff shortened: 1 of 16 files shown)


timonmerk (approved):
I agree with those changes. In the beginning I had in mind that some features could be a combination of others. To be honest, that would still be relevant, for example the ratio between beta and gamma is very relevant for decoding movement. This is at the moment however not implemented, so if it makes working with the toolbox for GUI or multiprocessing puprposes easier, you can merge it.

toni-neurosc:
I see, then let's wait on it until the project develops in a way that we need to go one way or the other. There might be other ways to handle the interdependency as well. Right now this can wait. Just to know, which feature is currently calculating "ratio between beta and gamma"?

timonmerk:
Currently there is no feature that is doing that. But I've seen that many times that it was very relevant for decoding. For example Zixiao's paper used those: https://www.nature.com/articles/s41746-024-01115-7

PR #350, opened 2024-06-26
Nm database

SamedVossberg:
- Integration of nm_database (new module)
- changes in run method -> can now receive flag "save_csv" to save full df as csv (otherwise only head)
- adjusted examples (save csv in run methods as I think it makes sense for the examples if people can look into the full csv files)


py_neuromodulation/nm_database.py
```diff
@@ -0,0 +1,82 @@
...
+        for file_path in glob.glob(pattern):
+            if os.path.exists(file_path):
+                if (self.db_time_idx/1000) - os.path.getctime(file_path) > 600:
+                    os.remove(file_path)
+
+        self.conn = sqlite3.connect(self.db_path, isolation_level=None)
+        self.cursor = self.conn.cursor()
+
+        db_dir = Path(folder_name)
+        if os.path.exists(db_dir):
```
timonmerk, line 28: Check if necessary

py_neuromodulation/nm_stream_offline.py
```diff
@@ -133,23 +138,35 @@ def _run(
...
-                feature_dict["time"] = np.ceil(time_[-1] * 1000 +1 ).astype(int)
+                feature_dict["time"] = np.ceil(time_[-1] * 1000 +1 )
                 logger.info("Time: %.2f", feature_dict["time"]/1000)
             
 
             self._add_target(feature_dict, data_batch)
+            buff_cnt += 1
+
+            feature_dict = db.cast_values(feature_dict)
+            db.create_table(feature_dict)
```
timonmerk, line 157: Move up above while loop

SamedVossberg: It has to stay in the while loop to init the table with the values from the loop. 
Solved this by putting it in an if statement

py_neuromodulation/nm_database.py
SamedVossberg: Indeed wasn't necessary in the latest version. Thanks!

py_neuromodulation/nm_database.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Class to create a database and insert data into it.
+    Parameters
+    ----------
+    out_path_root : str
+        The root path to save the database.
+    folder_name : str
+        The folder name to save the database.
+    """
+    def __init__(self, out_path_root, folder_name, csv_path = None):
```
timonmerk, line 17: Please add also typehints in the function calls for each parameter

py_neuromodulation/nm_database.py
```diff
@@ -0,0 +1,100 @@
...
+        self.out_path_root = out_path_root
+        self.folder_name = folder_name
+        self.db_path = Path(out_path_root, folder_name, "stream.db") 
+        if csv_path is None:
+            self.csv_path = Path(out_path_root, folder_name, f"stream.csv")
+        else:
+            self.csv_path = Path(csv_path)
+
+        if os.path.exists(self.db_path):
```
timonmerk, line 26: Is this required? So it cannot append data or overwrite?

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +139,37 @@ def _run(
...
 
-            l_features.append(feature_dict)
+            feature_dict = db.cast_values(feature_dict)
+            if not data_acquired:
+                db.create_table(feature_dict)
+                data_acquired = True
+            db.insert_data(feature_dict)
 
-        feature_df = pd.DataFrame(l_features)
+            if buff_cnt >= 10:
```
timonmerk, line 163: Please make the number of samples a function parameter. It would be even better if it's actually a time. So save data each 1 sec, 10 sec etc.

tests/test_all_features.py
```diff
@@ -27,7 +27,7 @@ def test_all_features_random_array():
     arr = np.random.random([2, 2000])
     stream = get_example_stream(arr)
     
-    df = stream.run(arr)
+    df = stream.run(arr, out_path_root="./test_data", folder_name="test_all_features_random_array")
```
timonmerk, line 30: The `./test_date` passes under windows?

timonmerk (changes requested):
Thanks for those changes @SamedVossberg. I have a few more code requests related to typehints in the function declarations, and making the number of samples / time interval between saves a function parameter.

py_neuromodulation/nm_database.py
SamedVossberg: I think appending isn't really an option here, is it? 
By default appending data would mean the db would grow infinitely until deleted. This could lead to errors, especially when working with different settings/features.

Overwriting is basically what we're doing here, as to my understanding when you overwrite you delete a file and create it again. Not deleting the db before will cause errors with column names.

SamedVossberg: And "normal" appending data during the stream (after db init) works by calling the ```insert_data``` method.

py_neuromodulation/nm_database.py
SamedVossberg: Good point, thanks. I added the typehints.

py_neuromodulation/nm_stream.py
SamedVossberg: So I added it as a parameter. I can easily change it to seconds instead of number of samples but I actually think number of feature samples would be more intuitive as users will know how many datapoints are possibly not written in the db in case of e.g. a crash. With seconds that could differ depending on the sampling rate, right? Also, when running a stream with a raw_data_generator time could be counter intuitive as there is the "real time" and the stream time (most likely running way faster). What do you think?

py_neuromodulation/nm_database.py
SamedVossberg: Yes, exactly. But that's also what it was like with the csv's to my understanding.

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
                 data_batch.astype(np.float64)
             )
             if is_stream_lsl:
-                feature_dict["time"] = time_[-1]
                 if self.verbose:
```
toniESAT, line 143: Not sure about this calculation of the batch time. I think that what should be being saved is either the first time of the batch or the last, not the length of the data. Also I think it's not correct that the first batch does not show the seconds of data processed. I attempted to fix it in my commit.

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
...
                     if last_time is not None:
                         logger.debug("%.3f seconds of new data processed", time_[-1] - last_time)
+                        feature_dict["time"] = time_[-1] - last_time
+                    else:
+                        feature_dict["time"] = 0
                     last_time = time_[-1]
```
toni-neurosc, line 149: I think the recording of the time is not quite right here. The logger should be outputting the seconds processed for the first batch, and the recorded time should be either the start or the end of the batch, not the length of it. I have attempted a fix in my commit, check it out.

py_neuromodulation/nm_stream_abc.py
```diff
@@ -155,6 +156,9 @@ def save_after_stream(
 
         self.save_sidecar(out_path_root, folder_name)
 
+        if not save_csv:
+            feature_arr = feature_arr.head()
```
toni-neurosc, line 160: I think it's better to not load the whole dataframe and subset it here, we can read just the first line from the SQLite database and pass it to this function

tests/test_all_features.py
timonmerk: Tests passed, so yes

py_neuromodulation/nm_stream.py
timonmerk: Agree

py_neuromodulation/nm_stream.py
timonmerk: Also, did you already push the commit?

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
                 data_batch.astype(np.float64)
```
toni-neurosc: I realized we don't need this, mne_lsl returns the data in `np.float64` format and `astype` actually makes a whole copy of the data, which is not good for performance

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
...
             )
             if is_stream_lsl:
-                feature_dict["time"] = time_[-1]
                 if self.verbose:
                     if last_time is not None:
                         logger.debug("%.3f seconds of new data processed", time_[-1] - last_time)
```
toni-neurosc: I don't think DEBUG level messages should depend on verbose, since DEBUG usually is disabled by default unless you set the level of logging  to DEBUG explicitely

py_neuromodulation/nm_stream.py
```diff
@@ -117,42 +129,58 @@ def _run(
...
-        l_features: list[dict] = []
-        last_time = None
+        prev_batch_end = 0
 
         while True:
             next_item = next(generator, None)
 
             if next_item is not None:
-                time_, data_batch = next_item
+                timestamps, data_batch = next_item
```
toni-neurosc, line 138: I changed this from time_ to timestamps, although I'm not sure is consistent with RawGenerator at this point

py_neuromodulation/nm_stream.py
```diff
@@ -117,42 +129,58 @@ def _run(
...
-                        logger.debug("%.3f seconds of new data processed", time_[-1] - last_time)
-                    last_time = time_[-1]
-            else:
-                feature_dict["time"] = np.ceil(time_[-1] * 1000 +1 ).astype(int)
-                logger.info("Time: %.2f", feature_dict["time"]/1000)
-            
+
+            feature_dict["time"] = (
+                batch_length if is_stream_lsl else np.ceil(this_batch_end * 1000 + 1)
+            )
```
toni-neurosc, line 155: I'm not entirely sure this is right for 2 reasons:
 - the "time" field should be consistent between file streams and lsl_streams, not having a duration for one and a time point for the other
 - Not sure if the units are consistent between both types of streams right now

toni-neurosc (commented):
So I have made a bunch of changes: 
- Got rid of the `out_path_root` and `folder_name` in all the internal functions, and handled everything at the Stream.run level.
- Removed the cast_values function from the database class as I don't think it belongs there, I just inlined it in the stream.run method
- I made some optimizations to the database after looking some articles on SQLite optimization, mainly to make sure we're going fast but also not losing data in case of crashes or system failures
- Moved the create_table function inside insert_data and the corresponding flag to the database class to not have internal database state mixed up with the stream class
- Added a head() method to the database class to read only 1 line from the database and avoid reading all the data to save just the header.
- Changed the code relating to the timestamps, more details in the review comments

I'm fine when it comes to the GUI side, I will merge this into the `gui` branch and if required make changes there. 

Btw, I made some tests fail somehow, will commit the fixes later.

I also have a big refactoring in the works that I almost finished when I was in Berlin, I'll try to finish that ASAP and make a PR because I don't want the GUI branch diverging too much from main either.

toni-neurosc (approved):

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
...
-        self.save_after_stream(out_path_root, folder_name, feature_df)
+        db.commit()
+        feature_df = db.fetch_all()
+
+
+        self.save_after_stream(out_path_root, folder_name, feature_df, save_csv=save_csv)
+
+        db.close()
 
         return feature_df
```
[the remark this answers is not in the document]
timonmerk: Yes, but I think it's still nice interface to kind of see the structure what features were calculated for which channels

py_neuromodulation/nm_database.py
```diff
@@ -0,0 +1,107 @@
...
+            os.remove(self.db_path)
```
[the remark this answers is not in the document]
timonmerk: Very elegant solution with the incremental file names!

py_neuromodulation/nm_stream.py
```diff
@@ -134,23 +140,38 @@ def _run(
...
-                feature_dict["time"] = np.ceil(time_[-1] * 1000 +1 ).astype(int)
+                feature_dict["time"] = np.ceil(time_[-1] * 1000 +1 )
                 logger.info("Time: %.2f", feature_dict["time"]/1000)
             
 
             self._add_target(feature_dict, data_batch)
+            buff_cnt += 1
 
-            l_features.append(feature_dict)
+            feature_dict = db.cast_values(feature_dict)
```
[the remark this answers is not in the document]
timonmerk: Agree

py_neuromodulation/nm_stream.py
timonmerk: Good catch

py_neuromodulation/nm_stream.py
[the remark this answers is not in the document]
toni-neurosc: The problem is, for long recordings, the moment the recording finishes it will try to load the entire database in memory and if it's too big it will hand or crash. Luckily thanks to the db itself this wouldn't cause a data-loss, but it's kinda awkward. For now I have added a `return_df` parameter True by default.

PR #357, opened 2024-08-30
add number of peaks features to sharpwaves

timonmerk:
request by @ashleywalton121


timonmerk:
Here the updated example repo @ashleywalton121 https://github.com/timonmerk/rns_waveformshape

PR #358, opened 2024-08-31
Add backend stream init and frontend setup page

timonmerk:


timonmerk:
Btw, I had to add this hard-coded textfield to get the path when reading the offline data file. Maybe you know more about it, or another way, but it seems through js it's not that easy to retrieve an absolute path from the `input` field: https://stackoverflow.com/questions/15201071/how-to-get-full-path-of-selected-file-on-change-of-input-type-file-using-jav 
The currently work-around is not optimal but maybe the only way to go.

Also tagging @SamedVossberg here to notify that I've worked on this branch. I think it shouldn't be too difficult to write a function to retrieve the nm_channels from `py_nm_state` backend.
By launching the `start_LSL_stream.py` file we can check the LSL stream setup.

SamedVossberg:
@timonmerk As soon as we merged the two PR's (this one and the channels) I will check if everything works as planned with the Channels and LSL as well 👍🏽

timonmerk:
@SamedVossberg Many thanks for your updates in the channel branch. I merged them in this branch for now.
I think it would at some point be great to have the whole website scrollable, also because some settings might extend vertically and we would loose them unless the zoom is used. But we can work on it later.

PR #359, opened 2024-09-05
Gui channels

SamedVossberg:
- Channels now fully connected to the backend
- some minor bugs

I saw there's another PR, so I don't know which one we want to merge first.
No conflicts here.

### DIFF
--- a/gui_dev/.gitignore
+++ b/gui_dev/.gitignore
@@ -12,3 +12,6 @@ node_modules
 dist
 dist-ssr
 *.local
+
+# Dev-scripts
+gui_startup.py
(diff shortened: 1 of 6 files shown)


timonmerk:
Many thanks @SamedVossberg! I tested your updated branch, and it works very well! I think one nice thing would be to make the page "scrollable", since this is currently lost with too many channel rows.

We now need to bring changes together. Currently we have the

- `gui_channels` branch including the updated channel backend handling
- `gui_lsl` with the LSL stream changes and stream setup page
- `liz_gui` branch including the updated (but not finalized settings page). @elizaveta-terek please make a pull request from the `liz_gui` branch into the `gui` branch. In this way you can get credits for your work and show potentially in applications about your work. Please confirm that you'll do it, otherwise we'll go ahead and make the PR ourself, since we need to merge the different branches soon to not diverge too much.

I think we should try to merge all of them into the `gui` branch.

@toni-neurosc I guess it makes sense that you finalize the settings, in the meantime I will try to merge Sam's and my branch.

SamedVossberg:
@timonmerk very true. I will add the scroll until tomorrow.

SamedVossberg:
So to make it scrollable there are basically two options:
- Making either the whole page scrollable or 
- Just having the ChannelsTable component scrollable
I opted for the latter as it seemed to me more user friendly having the possibility to click the button at the bottom at all time. Let me know what you think or if you would like to have anything changed @timonmerk.

timonmerk:
The branch has been merged into `gui_lsl`. Many thanks @SamedVossberg!

PR #360, opened 2024-09-10
Liz gui

elizaveta-terek:
This PR adds the basic setting components for different parameter types. 
Added: 
- Switch 
- Collapsible Box 
- Drag and Drop list 
- Dropdown 
- TextField 
- Frequency Settings 
The latter two need adjustments, as they are based on settings object flattening and do not update properly. Still, the Frequency Settings display a convenient layout of band selection, addition and edit. 
Switches allow to display feature settings themselves and later render the appearance of additional options, based on selected features. 
Collapsible boxes allow to separate the contents of several setting types and increase the readability of the code. 
Drag and Drop reflect the selected preprocessing options in the order they were selected. They can be later retracted from the zustand store.


toni-neurosc:
Hi @timonmerk , is it ok to you if i finish the Settings component here and merge directly to the main `gui` branch?

PR #365, opened 2024-10-02
Update bispectra feature

tsbinns:
Reduces some jankiness with the bispectra feature now that PyBispectra v1.2 is available (was already pinned in #355).

### DIFF
--- a/py_neuromodulation/features/bispectra.py
+++ b/py_neuromodulation/features/bispectra.py
@@ -96,11 +96,8 @@ class Bispectra(NMFeature):
     def calc_feature(self, data: np.ndarray) -> dict:
         from pybispectra import compute_fft, WaveShape
 
-        # PyBispectra's compute_fft uses PQDM to parallelize the calculation per channel
-        # Is this necessary? Maybe the overhead of parallelization is not worth it
-        # considering that we incur in it once per batch of data
         fft_coeffs, freqs = compute_fft(
-            data=np.expand_dims(data, axis=(0)),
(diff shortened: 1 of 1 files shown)


py_neuromodulation/features/bispectra.py
```diff
@@ -96,11 +96,8 @@ def __init__(
     def calc_feature(self, data: np.ndarray) -> dict:
         from pybispectra import compute_fft, WaveShape
 
-        # PyBispectra's compute_fft uses PQDM to parallelize the calculation per channel
-        # Is this necessary? Maybe the overhead of parallelization is not worth it
-        # considering that we incur in it once per batch of data
         fft_coeffs, freqs = compute_fft(
```
tsbinns, line 99: I switched away from `pqdm`, so this comment is no longer true.

py_neuromodulation/features/bispectra.py
```diff
@@ -127,12 +124,11 @@ def calc_feature(self, data: np.ndarray) -> dict:
             f1s=tuple(self.settings.f1s),  # type: ignore
             f2s=tuple(self.settings.f2s),  # type: ignore
         )
+        waveshape = waveshape.results.get_results(copy=False)  # can overwrite obj with array
```
tsbinns, line 127: This is a less hacky way of extracting the bispectral info while avoiding a copy, without relying on private attrs.

timonmerk:
Many thanks for the contribution @tsbinns!

PR #366, opened 2024-10-02
Fix icoh normalisation term and add unit test

tsbinns:
Two things in this PR:
- Adds a unit test for the coherence feature, including a sanity check that the connectivity values lie in an expected range using simulated connectivity data.
- Fixes a bug for icoh where it was not being normalised properly, causing connectivity values to not lie in the expect [-1, 1] range.


py_neuromodulation/features/coherence.py
```diff
@@ -83,10 +83,14 @@ def get_coh(self, feature_results, x, y):
         self.Pyy = welch(y, self.sfreq, self.window, nperseg=128)[1]
         self.Pxy = csd(x, y, self.sfreq, self.window, nperseg=128)[1]
 
-        if self.coh:
-            self.coh_val = np.abs(self.Pxy**2) / (self.Pxx * self.Pyy)
-        if self.icoh:
-            self.icoh_val = np.array(self.Pxy / (self.Pxx * self.Pyy)).imag
```
tsbinns, line 89: Denominator term should have the square root taken. Unfortunately the strategy for coh where you square the numerator and take the absolute value doesn't work for icoh since we extract only the imaginary part and this info isn't properly preserved.

py_neuromodulation/features/coherence.py
```diff
@@ -83,10 +83,14 @@ def get_coh(self, feature_results, x, y):
...
+        if self.coh and self.icoh:
+            cohy = self.Pxy / np.sqrt(self.Pxx * self.Pyy)
+            self.coh_val = np.abs(cohy)
+            self.icoh_val = cohy.imag
+        elif self.coh:
+            self.coh_val = np.abs(self.Pxy ** 2) / (self.Pxx * self.Pyy)
+        else:  # only self.icoh == True
+            self.icoh_val = self.Pxy.imag / np.sqrt(self.Pxx * self.Pyy)
```
tsbinns, line 93: Just a suggestion, but if both coh and icoh are requested features, then I expect it is faster to compute coherency once, then extract coh and icoh values. Otherwise if only one is requested, it can follow the same path as before of computing this information separately.

tsbinns:
Another note, I would argue it is beneficial to expose the `nperseg` parameter. With a high sampling frequency like 500 Hz, `nperseg=128` is completely fine, but as sampling frequency decreases, this can come very close to the max possible. The result is coherence values which converge towards 1 (i.e. perfect correlation) regardless of the data passed in, and icoh values which are completely erratic.

At least if the end-user has control over `nperseg`, they can reduce this to reflect their sampling frequency. Doing so decreases the spectral resolution, but at least the connectivity values are still valid.

https://github.com/neuromodulation/py_neuromodulation/blob/12d3ed967b69dcf81550e8e069cec7e414f610a9/py_neuromodulation/features/coherence.py#L82-L84

What are people's thoughts on exposing this param as a setting?

tests/test_coherence.py
```diff
@@ -0,0 +1,122 @@
...
+    settings.coherence_settings.frequency_bands = ["signal", "noise_low", "noise_high"]
+
+    # only average within each band required
+    settings.coherence_settings.features = {
+        "mean_fband": True, "max_fband": False, "max_allfbands": False
+    }
+
+    # connectivity indices, i.e.: ([0, 1], [1, 0])
+    # FIXME: indices with only 1 channel in seeds/targets raises error with type validatn
+    settings.coherence_settings.channels = [ch_names, ch_names[::-1]]
```
tsbinns, line 73: One thing I came across was a `ValidationError` from `pydantic` if I tried to pass in indices that contained lists of only one entry, e.g. `[["seed"], ["target"]]`:
```
List should have at least 2 items after validation, not 1 [type=too_short, input_value=['seed'], input_type=list]
```

This means you have to always look at connectivity between more than just 2 channels.

A workaround is to do what I did here and just repeat the indices, e.g. `[["seed", "target"], ["target", "seed"]]`, but that involves unnecessarily computing an extra connection.

I'm not familiar with `pydantic`; is it possible to ignore/whitelist the check for this case?

timonmerk:
Also @tsbinns, it definitely makes sense to expose `nperseg`. Tbh I did not even think about that, but especially for lower frequencies that will be necessary!

PR #371, opened 2024-10-05
Gui graphs

timonmerk:
@SamedVossberg I've created this PR so some changes can be discussed


timonmerk:
@SamedVossberg In the `mnelsl_stream.py` you wrote previously this code:

 

>                 for i in range(self._n_seconds_wait_before_disconnect):
>                     if (
>                         data is not None
>                         and check_data is not None
>                         and np.allclose(data, check_data, atol=1e-7, rtol=1e-7)
>                     ):
>                         logger.warning(
>                             f"No new data incoming. Disconnecting stream in {3-i} seconds."
>                         )
>                         time.sleep(1)
>                         i += 1
>                         if i == self._n_seconds_wait_before_disconnect:
>                             self.stream.disconnect()
>                             logger.warning("Stream disconnected.")
>                             break

What was the motivation behind that? 

I guess something related to the example test run? I removed it now so that the LSL Stream can be tested in the frontend

SamedVossberg:
@timonmerk The idea was to have a timeout for the lsl stream to terminate if no new data was incoming. As back then there was no option to stop the stream manually (we tried with a keyboard interrupt first back then but then decided against it) the timeout terminated the lsl connection if there wasn't any new data incoming for three consecutive seconds. And yes, it can be deleted as with the frontend established this will be deprecated.

timonmerk:
@SamedVossberg I updated now the files that the full spectrums are returned in the features. Unfortunately though there is a bug in the settings that I am not able to fix, as mentioned here: https://github.com/neuromodulation/py_neuromodulation/issues/377#issuecomment-2461908848

@toni-neurosc it would be great if you could have a look how we could find a simple solution how to fix this bug. I tried to remove all the `preprocessing` elements, that it's an empty list, but it still kept giving me the bug.
Unfortunately I don't have more time at the moment to fix this point, but it would be great in order to continue testing the GUI.

Also, regarding the issue there we exceed the number of columns in the SQL database, I would suggest for now to limit the number of channels, in this way the feature number doesn't exceed the 2000 sqlite limit. Not a nice solution.. but something that allows us to keep developing at the moment.

timonmerk:
@SamedVossberg would it be possible to merge the changes from your `gui_data_processor` branch in this one? For me it's mainly about the parts that fix the HeatMap visualization. I saw that you wrote some rust parts also, do you think we should merge those also in the gui branch?

PR #380, opened 2024-11-13
Fixed use of non-target/decoding channel in the combined setting

alessiaca:
For visualization purposes I included an extra channel in the feature array, which was neither a label nor a used channel. However, it was still included in the combined data array and thus, for decoding. The proposed change makes sure that only those channels that are also present in the used_ch list (and not all apart from the label and the time) are included in the combined dataset.


timonmerk:
Thanks a lot @alessiaca!

PR #382, opened 2025-01-24
GUI

timonmerk:


timonmerk:
@toni-neurosc, I am trying to wrap up the merge of this branch into main. I merged now the main branch into `gui`, and fixed the unit tests. There are two main issues remaining:
1. `plot_7_lsl_example.py` passes through the whole code, but doesn't finish executing. In debug mode it ends up in a `QueueFeederThread` which doesn't stop. I tried to close all `multiprocessing.queues`, and also call `LSLPlayer.stop()`, but it doesn't fix the execution..  
2. Due to the recent GUI setting changes, the `plot_2_example_add_feature.py` example got broken. The function `nm.add_custom_feature("channel_mean", ChannelMean)` does not seem to work, when initializing the settings: `settings = nm.NMSettings()` the following exception comes up:

> Exception has occurred: ValueError
> "FeatureSelector" object has no field "channel_mean"
>   File "/Users/Timon/Documents/py_neuromodulation/py_neuromodulation/stream/settings.py", line 133, in __init__
>     setattr(self.features, feat_name, True)
>   File "/Users/Timon/Documents/py_neuromodulation/examples/plot_2_example_add_feature.py", line 62, in <module>
>     settings = nm.NMSettings() # Get default settings
>                ^^^^^^^^^^^^^^^
> ValueError: "FeatureSelector" object has no field "channel_mean"

 I am not sure if `setattr` works with pydantic objects? Maybe this is the underlying issue.

Additionally 3. the execution time of real-time streaming in the GUI is still extremely slow, and the websocket breaks. @toni-neurosc if you found a fix for that issue it would be really great if you could push it here. Many thanks!

toni-neurosc:
Hi Timon, fixed issues 1 and 2 with my recent commits. Short summary

1. The backend interface class should not be used by the end users, only by the backend server for the GUI. Maybe it was a bad idea to put it as a parameter in the "run" function, we should remove it, change it to a "private" variable (with a leading _ ) and just set it manually in the PyNMState class

2. This was just a custom settings in the Pydantic model NMBaseModel using the ConfigDict Pydantic class that was removed at some point during development of the GUI branch, it's now back in and working properly

For number 3, I need a bit more time since I managed to fix React re-rendering the whole page on every plot update, but for now it's just a proof of concept in the power spectrum graph, I need to re-work the socket store a bit to:
- Centralize the separation of the data points for each channel/feature in the socket store to prevent redundant work in each of the graph components
- Extend the Zustand suscription model to the rest of the graphs and maybe centralize the subscription logic into some kind of graph "superclass" that can be instantiated per graph "style" or sth.

Also, I haven't fully grasped the Zustand subscribe thing fully yet, I am having an issue where on restarting the stream the graph stops listening to updates, so I need to figure that out. But early results are very promising as we went from updating each and everyone of the components in the page (including every button, dropdown, everything) to just the graphs.

I have published a branch with the changes to BandPowerGraph.jsx if you want to see how it works.

PR #388, opened 2025-03-19
[Performance] Raw Graph Performance Improvement

ahmedtarek-:
Still WIP


ahmedtarek-:
## Update

**Debugging performance for 500 messages**
    - `.Restyle()` yTraces only -> `611 update, 121.921 s`
    - `.React()` with subscription only -> `500 updates, 114.072 s`
    - `.React()` without subscription (before this PR) -> `254 updates, 22.481 s`

This basically means that the subscription direction isn't performing well. The only idea I have in mind now is to batch the data/messages and update the graph every x amount of messages. Apart from this, I ran out of ideas how to improve the subscription logic.

ahmedtarek-:
Okk, will do the ECharts direction then ✌🏽

PR #390, opened 2025-05-05
[Performance] Using Echarts To Improve Raw Graph Performance

ahmedtarek-:
WIP


ahmedtarek-:
[screen-capture.webm](https://github.com/user-attachments/assets/3906da1e-5e52-404b-a313-da8d3f0e77a6)
I could only make it work, after being stuck for some time trying to make it work with subscription logic, as shown in the video.
It looks ugly af at the moment. On the bright side, the performance seems to be better and not lagging behind (the video is taken after running it for more than 2 minutes already).

My current task is to
1. Make it prettier; more spaces between lines
2. Quantify the performance and compare to the other pull request

timonmerk:
Closed for now. Might check up later

Issue #405, opened 2025-08-01
Add tests with zeros


timonmerk:
Fooof fails wtih all zeros; add exception

Issue #402, opened 2025-07-31
Fix "average" rereference if not in list


timonmerk:
Currently the typechecks fail if "average" is passed in channel init as a string

PR #406, opened 2025-08-07
Hotfix fooof channel reref

timonmerk:
addresses #402, #403, #405


timonmerk:
TODO: Write unit-test for bad channel

